PR #35, opened 2023-07-11
Add extra FAQ entry

patrick-kidger:


packquickly:
Should we have further elaboration here as to why we didn't choose to include these?

patrick-kidger:
No strong feelings either way, really.

PR #56, opened 2023-10-31
Improved RESULT ennum successful string

### LINKED ISSUE #55: Ambiguous output from successful solution using `lineax.RESULTS`
I originally encountered this issue in `optimistix` but realised it originates here. 

After a successful solution is found the `solution.result` value is `0` when it has succeeded, with a series of other values for the various failure modes, all fine so far. However following the docs to get a 'human-readable' interpretation using `lineax.RESULTS[sol.result]` returns an empty string when it has succeeded. As far as I can tell without knowledge of `Enumeration` objects or looking at the code directly there is really no way to know that is actually a good thing, rather than some form of 'null result' as I originally interpreted it. 

Maybe this is intentional for some deeper reason, but not as far as I can see. I think having the success message print out something as simple a `'success'` would go a long way for new users. 

Let me know if you're interested and Ill make a PR for this.

LouisDesdoigts:
Closes #55

### DIFF
--- a/lineax/_solution.py
+++ b/lineax/_solution.py
@@ -41,7 +41,7 @@ If you *were* expecting this solver to work with this operator, then it may be b
 
 
 class RESULTS(eqxi.Enumeration):
-    successful = ""
+    successful = "Success."
     max_steps_reached = (
         "The maximum number of solver steps was reached. Try increasing `max_steps`."
     )


patrick-kidger:
I'm not sure what's causing the test failure, as I can't reproduce it locally. (Even setting the same seed etc.) So I think we can ignore that for now.

I have however realised that this change will break Optimistix: right now, `eqx.Enumeration`s check that you don't change a message when subclassing, and now `optimistix.RESULTS.successful` will try to change it back to an empty string. (As `optimistix.RESULTS` subclasses `lineax.RESULTS`.)

I'm going to downgrade that to a warning. Once the next Equinox release is out, I'll merge this.

patrick-kidger:
Coming back around, I'm afraid that I'm going to close this as a 'wontfix'. :(
It ends up being a little bit annoying to change this, and then propagate the changed result across the ecosystem, in a way that won't spam everyone with errors and warnings. Doable, but probably not worth the hassle.

Thank you anyway!

PR #58, opened 2023-11-01
[WIP] Complex dtype support

Randl:
Fix https://github.com/google/lineax/issues/57

- [ ] Add `dtype` parameter to tests, test with both real and complex number
  - [ ] make sure data types are explicit in all tests
- [x] Change typing from `Float` to `Inexact`
- [x] Add `holomorphic` to appropriate places in `jaxfwd`
- [ ] Add complex-related tags and methods, e.g., `adjoint` (as opposed to `transpose`)
- [ ] Fix complex-specific bugs in methods, e.g., missing conjugates, make sure tests are passing
- [ ] Add complex-specific and new functionality (such as adjoint) tests

### DIFF
--- a/lineax/__init__.py
+++ b/lineax/__init__.py
@@ -66,6 +66,7 @@ from ._tags import (
     lower_triangular_tag as lower_triangular_tag,
     negative_semidefinite_tag as negative_semidefinite_tag,
     positive_semidefinite_tag as positive_semidefinite_tag,
+    self_adjoint_tag as self_adjoint_tag,
     symmetric_tag as symmetric_tag,
     transpose_tags as transpose_tags,
     transpose_tags_rules as transpose_tags_rules,
--- a/lineax/_misc.py
+++ b/lineax/_misc.py
@@ -90,11 +90,11 @@ class NoneAux(eqx.Module):
         return self.fn(*args, **kwargs), None
 
 
-def jacobian(fn, in_size, out_size, has_aux=False):
+def jacobian(fn, in_size, out_size, holomorphic=False, has_aux=False):
     # Heuristic for which is better in each case
     # These could probably be tuned a lot more.
     if (in_size < 100) or (in_size <= 1.5 * out_size):
-        return jax.jacfwd(fn, has_aux=has_aux)
+        return jax.jacfwd(fn, holomorphic=holomorphic, has_aux=has_aux)
     else:
         return jax.jacrev(fn, has_aux=has_aux)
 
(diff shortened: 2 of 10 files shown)


Randl:
If I put float64/complex128 in tests, most of them pass, except for a number of jvp tests wtih singular complex matrices for SVD and some transpose tests. However, with float32/complex64, additional tests fail (due to lack of precision?). So three main questions are which types to use in tests to make sure that everything works, which additional tests are needed specifically for complex variables, and what is the problem with SVD.
@patrick-kidger any insights?

patrick-kidger:
Thanks for starting in on this! This is really cool to see.

Regarding the float32 tests, this is expected. We actually set float64 as the default for any test which doesn't already specify it explicitly: https://github.com/google/lineax/blob/9a573529f3eea2a6944009f6e4afe4f364366baf/tests/conftest.py#L24
It's fairly common for linear solves to be a bit finickity on this front.  
So likewise, I'd be happy to use just complex128 in the tests.

Regarding "which additional tests are needed specifically for complex variables". A couple of things come to mind:
- That `_misc.two_norm` and `_misc.max_norm` are adjusted to do the right thing, along with their JVPs. This should actually be fairly easy as I actually did this in Optimistix [here](https://github.com/patrick-kidger/optimistix/blob/4e97e487ce30506e34744bbb68423b7c5f330180/optimistix/_misc.py#L126-L198) with associated tests [here](https://github.com/patrick-kidger/optimistix/blob/4e97e487ce30506e34744bbb68423b7c5f330180/tests/test_misc.py#L39-L91). I think we just need to upstream the implementation to Lineax. (And side note, probably then make them available under `lineax.internal.{two,max}_norm` so we can just have Optimistix depend on the implementation here.)
- That we get transposes, not conjugate-transposes, when performing `jax.linear_transpose` of a `lineax.linear_solve`. (This is the convention JAX uses here for complex numbers.)

This is in addition to the tests for "getting the right forward solution" and "getting the right gradient" etc., which can be handled by extending existing tests.

---

You note that there's a few things that need adjusting. I'd ask that we do each piece across separate PRs, to make it easier for me to review them.

Randl:
@patrick-kidger I've split the changes in 5 PRs, which I believe should be non-breaking if merged in order (the dtype one should be non-breaking by itself). 
Since the changes are not complete yet, these PRs do not fully cover the changes; in particular, the tests for complex dtype are still off.

patrick-kidger:
Great, thank you -- I've gone through and left a few comments in some PRs. Overall I like the look of these changes!

patrick-kidger:
Closing in favour of the individual PRs.

PR #60, opened 2023-11-01
Add holomorphic parameter

Randl:
Add option to pass `holomorphic` to `jacobian` function, so that it will work with complex functions


lineax/_misc.py
```diff
@@ -90,11 +90,11 @@ def __call__(self, *args, **kwargs):
         return self.fn(*args, **kwargs), None
 
 
-def jacobian(fn, in_size, out_size, has_aux=False):
+def jacobian(fn, in_size, out_size, holomorphic=False, has_aux=False):
```
patrick-kidger, line 93: This needs piping into the `jacrev` branch as well.

Randl: Looks like this branch is untested. Worth adding testcase somewhere

PR #61, opened 2023-11-01
Fixes for complex-related changes (mostly conjugates)

Randl:
Test depends on other PRs

### DIFF
--- a/lineax/_solver/gmres.py
+++ b/lineax/_solver/gmres.py
@@ -257,7 +257,11 @@ class GMRES(AbstractLinearSolver[_GMRESState]):
                 lambda x: jnp.pad(x[..., None], ((0, 0),) * x.ndim + ((0, restart),)),
                 r_normalised,
             )
-            coeff_mat_init = jnp.eye(restart, restart + 1, dtype=r_normalised[0].dtype)
+            coeff_mat_init = jnp.eye(
+                restart,
+                restart + 1,
+                dtype=jnp.result_type(*jtu.tree_leaves(r_normalised)),
+            )
 
             def cond_fun(carry):
                 _, _, breakdown, step = carry
--- a/lineax/_solver/gmres.py
+++ b/lineax/_solver/gmres.py
@@ -257,7 +257,7 @@ class GMRES(AbstractLinearSolver[_GMRESState]):
                 lambda x: jnp.pad(x[..., None], ((0, 0),) * x.ndim + ((0, restart),)),
                 r_normalised,
             )
-            coeff_mat_init = jnp.eye(restart, restart + 1, dtype=r_norm.dtype)
+            coeff_mat_init = jnp.eye(restart, restart + 1, dtype=r_normalised[0].dtype)
 
             def cond_fun(carry):
                 _, _, breakdown, step = carry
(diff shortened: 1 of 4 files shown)


tests/test_solve.py
```diff
@@ -62,6 +62,19 @@ def test_mixed_dtypes(solver):
...
+    c128 = lambda x: jnp.array(x, dtype=jnp.complex128)
+    x = [[c64(1), c128(5.0j)], [c64(2.0j), c128(-2)]]
+    y = [c128(3), c128(4)]
+    struct = jax.eval_shape(lambda: y)
+    operator = lx.PyTreeLinearOperator(x, struct)
+    out = lx.linear_solve(operator, y, solver=solver).value
+    true_out = [c64(-0.75 - 2.5j), c128(0.5 - 0.75j)]
+    assert shaped_allclose(out, true_out)
+
+
```
patrick-kidger, line 77: Actually, one other thing we should test here is mixed float64+complex128.

lineax/_solver/gmres.py
```diff
@@ -257,7 +257,7 @@ def main_gmres(y):
                 lambda x: jnp.pad(x[..., None], ((0, 0),) * x.ndim + ((0, restart),)),
                 r_normalised,
             )
-            coeff_mat_init = jnp.eye(restart, restart + 1, dtype=r_norm.dtype)
+            coeff_mat_init = jnp.eye(restart, restart + 1, dtype=r_normalised[0].dtype)
```
patrick-kidger, line 260: I think this is wrong. `r_normalised` is an arbitrary pytree:
https://github.com/google/lineax/blob/26133fa1c92ad096c25a77ef5493a49bc8738332/lineax/_solver/gmres.py#L399
so it may not support indexing like this.

I'm also not sure whether these coefficients should ever be complex-valued, or if they should always be real-valued. @packquickly any thoughts?

patrick-kidger:
Alright, I've started to go through and merge things!

I think I'd probably suggesting merging this PR with #62, as both involve modifying the existing solvers and adding tests.

Thinking about tests: as per the discussion in https://github.com/google/lineax/pull/58#issuecomment-1789410622, I think we also need to extend some of the existing tests to also check complex numbers. In particular that every solver handles complex numbers, that JVPs work, and that transposes/grads work. Those are all places where things can easily go wrong.

lineax/_solver/gmres.py
Randl: I think it should be complex. It is a Gram–Schmidt process, and the coefficients in GS can be complex. Also, it fixed some tests :)
Should it be something like `jnp.result_type(jtu.tree_flatten(r_normalised)[0])` then? I'm not too proficient with PyTrees

Randl:
As for merging with #62, I thought splitting these two would make it easier to review, especially since #62 is basically a standalone feature implementation, while this is a set of fixes. I can merge them if you prefer.

As for tests, after these are merged, I plan to add `complex128` to the list of dtypes of tests; this will test what should already be mostly a working complex library and will allow us to hunt remaining problems. This also should help to understand which complex-specific tests should be added.

lineax/_solver/gmres.py
patrick-kidger: More or less, yup: `jnp.result_type(*jtu.tree_leaves(r_normalised))`, I think.

lineax/_solver/qr.py
```diff
@@ -79,7 +79,7 @@ def compute(
         else:
             # Least squares solution if overdetermined.
             solution = jsp.linalg.solve_triangular(
-                r, q.T @ vector, trans="N", unit_diagonal=False
+                r, q.T.conj() @ vector, trans="N", unit_diagonal=False
```
patrick-kidger, line 82: I suspect there should be a `conj` somewhere on the `if transpose` branch as well.

patrick-kidger:
Nit aside I think this LGTM!

patrick-kidger:
Alright, LGTM!

PR #63, opened 2023-11-01
Add explicity dtype parameter to tests

Randl:
This PR makes datatypes explicit, and will allow in future to test with more than one datatype

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -33,9 +33,9 @@ def getkey():
 
 
 @ft.lru_cache(maxsize=None)
-def _construct_matrix_impl(getkey, cond_cutoff, tags, size):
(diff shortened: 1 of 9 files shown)


tests/helpers.py
```diff
@@ -33,9 +33,9 @@ def getkey():
 
 
 @ft.lru_cache(maxsize=None)
-def _construct_matrix_impl(getkey, cond_cutoff, tags, size):
+def _construct_matrix_impl(getkey, cond_cutoff, tags, size, dtype=jnp.float64):
```
patrick-kidger, line 36: I'd suggest removing the default, and just explicitly specifying it everywhere. Default arguments are often a source of unexpected behaviour, so my usual rule is to avoid them across internal APIs.

Issue #57, opened 2023-11-01
Support complex dtypes


Randl:
Following discussion in https://github.com/patrick-kidger/diffrax/pull/326#discussion_r1378181082
From the first glance, the following is the minimum list of things to do for complex number support:
- [ ] Add `dtype` parameter to tests, test with both real and complex number
- [ ] Change typing from `Float` to `Inexact`
- [ ] Add `holomorphic` to appropriate places in `jaxfwd`
- [ ] Add complex-related tags and methods, e.g., `adjoint` (as opposed to `transpose`)
- [ ] Fix complex-specific bugs in methods, e.g., missing conjugates, make sure tests are passing
- [ ] Add complex-specific and new functionality (such as adjoint) tests 

I've tried to tackle part of the list at https://github.com/Randl/lineax/tree/complex_support
I will be happy to collaborate or further discuss the required steps.

patrick-kidger:
Agreed!
Seroiously, great job pushing this all the way through. I'm really happy to have this in -- and I've just pulled the trigger for the next release of Lineax!

PR #64, opened 2023-11-12
Enable complex tests

Randl:
Some tests are still failing, I'm currently checking it.

### DIFF
--- a/tests/test_adjoint.py
+++ b/tests/test_adjoint.py
@@ -14,7 +14,7 @@ from .helpers import (
 
 
 @pytest.mark.parametrize("make_operator", make_operators)
-@pytest.mark.parametrize("dtype", (jnp.float32, jnp.complex64))
+@pytest.mark.parametrize("dtype", (jnp.float64, jnp.complex128))
 def test_adjoint(make_operator, dtype, getkey):
     if make_operator is make_diagonal_operator:
         matrix = jnp.eye(4, dtype=dtype)
@@ -52,6 +52,16 @@ def test_functional_pytree_adjoint():
         return {"b": y["a"]}
 
     y_struct = jax.eval_shape(lambda: {"a": 0.0})
+    operator = FunctionLinearOperator(fn, y_struct)
+    conj_operator = lx.conj(operator)
+    assert tree_allclose(lx.materialise(conj_operator), lx.materialise(operator))
+
+
+def test_functional_pytree_adjoint_complex():
+    def fn(y):
+        return {"b": y["a"]}
+
+    y_struct = jax.eval_shape(lambda: {"a": 0.0j})
     operator = FunctionLinearOperator(fn, y_struct)
     conj_operator = lx.conj(operator)
     assert tree_allclose(lx.materialise(conj_operator), lx.materialise(operator))
--- a/tests/test_jvp_jvp.py
+++ b/tests/test_jvp_jvp.py
@@ -41,7 +41,7 @@ from .helpers import (
         construct_singular_matrix,
     ),
 )
-@pytest.mark.parametrize("dtype", (jnp.float64,))
+@pytest.mark.parametrize("dtype", (jnp.float64, jnp.complex128))
 def test_jvp_jvp(
     getkey, solver, tags, pseudoinverse, make_operator, use_state, make_matrix, dtype
 ):
--- a/tests/test_misc.py
+++ b/tests/test_misc.py
@@ -16,6 +16,7 @@ import jax
 import jax.numpy as jnp
 import lineax as lx
 import lineax._misc as lx_misc
+import pytest
 
 
 def test_inexact_asarray_no_copy():
@@ -32,7 +33,8 @@ def test_inexact_asarray_jvp():
     assert type(t) is not float
 
 
-def test_zero_matrix():
-    A = lx.MatrixLinearOperator(jnp.zeros((2, 2)))
-    b = jnp.array([1.0, 2.0])
+@pytest.mark.parametrize("dtype", (jnp.float64, jnp.complex128))
+def test_zero_matrix(dtype):
+    A = lx.MatrixLinearOperator(jnp.zeros((2, 2), dtype=dtype))
+    b = jnp.array([1.0, 2.0], dtype=dtype)
     lx.linear_solve(A, b, lx.SVD())
--- a/tests/test_operator.py
+++ b/tests/test_operator.py
@@ -111,28 +111,28 @@ def _assert_except_diag(cond_fun, operators, flip_cond):
             assert cond_fun(operator)
 
 
-@pytest.mark.parametrize("dtype", (jnp.float64,))
+@pytest.mark.parametrize("dtype", (jnp.float64, jnp.complex128))
 def test_linearise(dtype, getkey):
     operators = _setup(getkey, jr.normal(getkey(), (3, 3), dtype=dtype))
     for operator in operators:
         lx.linearise(operator)
 
 
-@pytest.mark.parametrize("dtype", (jnp.float64,))
+@pytest.mark.parametrize("dtype", (jnp.float64, jnp.complex128))
 def test_materialise(dtype, getkey):
     operators = _setup(getkey, jr.normal(getkey(), (3, 3), dtype=dtype))
     for operator in operators:
         lx.materialise(operator)
 
 
-@pytest.mark.parametrize("dtype", (jnp.float64,))
+@pytest.mark.parametrize("dtype", (jnp.float64, jnp.complex128))
 def test_materialise_large(dtype, getkey):
     operators = _setup(getkey, jr.normal(getkey(), (200, 500), dtype=dtype))
     for operator in operators:
         lx.materialise(operator)
 
 
-@pytest.mark.parametrize("dtype", (jnp.float64,))
+@pytest.mark.parametrize("dtype", (jnp.float64, jnp.complex128))
 def test_diagonal(dtype, getkey):
     matrix = jr.normal(getkey(), (3, 3), dtype=dtype)
     matrix_diag = jnp.diag(matrix)
@@ -141,7 +141,7 @@ def test_diagonal(dtype, getkey):
         assert jnp.allclose(lx.diagonal(operator), matrix_diag)
 
 
-@pytest.mark.parametrize("dtype", (jnp.float64,))
+@pytest.mark.parametrize("dtype", (jnp.float64, jnp.complex128))
 def test_is_symmetric(dtype, getkey):
     matrix = jr.normal(getkey(), (3, 3), dtype=dtype)
     symmetric_operators = _setup(getkey, matrix.T @ matrix, lx.symmetric_tag)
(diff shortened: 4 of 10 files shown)


Randl:
18 tests fail locally, all related to gradients for singular complex matrices (with different solvers). @patrick-kidger ideas?

patrick-kidger:
If I had to guess, the JVP rule we currently use is probably wrong for complex numbers. This is currently implemented [here](https://github.com/google/lineax/blob/49e83ef348378327ff2e6cbdb4f7e09a7bc7d83b/lineax/_solve.py#L129-L241), including a fairly detailed comment about the derivation.

Randl:
You're right. The complex formula has conjugate transpose instead of transpose (see, e.g., this [mathoverflow question](https://mathoverflow.net/questions/307717/derivative-of-complex-matrix-pseudo-inverse-with-respect-to-real-and-imaginary-c) and references there).

So, if I understand correctly, these three lines should all have conjugation on them
https://github.com/google/lineax/blob/49e83ef348378327ff2e6cbdb4f7e09a7bc7d83b/lineax/_solve.py#L193-L195
but we don't have a conjugation of the operator. In particular, if I manually change transpose in QR to conj transpose, some more tests pass, so I am fairly sure this is at least part of the required fix. What would be the cleanest way to do that?

patrick-kidger:
I think what you've done so far looks reasonable! The latest change for JVPs in particular.
I'm not sure what's going on with the currently-failing test, we'll have to dig into that.

Randl:
I'm pretty sure that the remaining test is failing because 
https://github.com/google/lineax/blob/49e83ef348378327ff2e6cbdb4f7e09a7bc7d83b/lineax/_solve.py#L195
does transpose:
https://github.com/google/lineax/blob/49e83ef348378327ff2e6cbdb4f7e09a7bc7d83b/lineax/_solver/qr.py#L87-L92
but we need conjugate transpose, i.e., something like
```python
state = ((q.conj(), r.conj()), not transpose, transposed_packed_structures,)
```
I'm not sure what would be the simplest way to conjugate solver?

Randl:
I think maybe we should just add `.conj()` to each solver, which is consistent with having `conj` for operators. I'd prefer to hear your opining before implementing it.

patrick-kidger:
Ah, that makes sense. Yes, I think that sounds like a good solution.

Randl:
Weird, the test passed three times and crashed once. I'll take a look on that

Randl:
The test fails randomly (as in can fail and pass with the same seed in getkey). Not sure what other sources of randomness are there in the tests?

patrick-kidger:
Oh that is weird!
Hmm, maybe something inside one of the calls JAX is making to LAPACK or something? I'm not sure.

Randl:
I've run more tests with `pytest-repeat`:
```
 % pytest tests/test_jvp_jvp.py --count=400 -v -k "test_jvp_jvp[complex128-construct_matrix-True-make_jac_operator-solver17-tags17-False" 
```
Looks like the test fails in ~2% of cases (I was either lucky or optimistic when tried to estimate earlier). The same test with float64 didn't fail once in 600+ runs.

Should we merge functional changes in a separate PR and leave this open until we figure it out?

patrick-kidger:
Have you tried adding `jax.debug.print` statements through the internals, and checking which operation is producing nondeterministic results? I think it should be fairly simple thing to bisect through the computation to track down where the issue is.

Randl:
I found another source of randomness that wasn't fixed (the helper used a separate key), but it doesn't seem to be all. I still can't get reproducible failure.

Randl:
I've left only `modes = (  {"vec"}, )` to iterate faster 

This line sometimes gives different results (seemingly with much larger magnitude):
https://github.com/google/lineax/blob/49e83ef348378327ff2e6cbdb4f7e09a7bc7d83b/lineax/_solver/cg.py#L175

e.g., typical run (seed 443228094), `mat_p = [-15.46805745+57.8129518j  -85.02231472+87.6522955j
  33.67970308+49.15732457j]`, failed run `mat_p = [-139.07700475+139.02408315j  -48.51520874+139.77350776j
   19.67462015 +39.8780054j ]`. Everything beforehand appears identical. Typical runs are consistent between them, i.e., running same seed 5 times will give same value, except if fails.

Randl:
It's normalCG with Jacobian operator. You can see the exact parameters in command I sent earlier.

Randl:
```
getkey = _GetKey(seed=301823979, call=9, key=Array([        0, 301823979], dtype=uint32))
solver = NormalCG(
  rtol=1e-12,
  atol=1e-12,
  norm=<wrapped function max_norm>,
  stabilise_every=10,
  max_steps=None
), tags = negative_semidefinite_tag
pseudoinverse = False, make_operator = <function make_jac_operator at 0x15b81a3e0>, use_state = True, make_matrix = <function construct_matrix at 0x15a483f60>
dtype = <class 'jax.numpy.complex128'>
```

patrick-kidger:
Gotcha. Hmm, looking at the implementation nothing jumps out as particularly weird:

https://github.com/google/lineax/blob/199c27777dea454f6a7d23c531664dba9bd2cbf9/tests/helpers.py#L215

Can we track the nondeterminism to a particular operation inside this? Assuming that we're producing the same random samples each run, then possibly this is an indication of a bug somewhere in JAX itself?

patrick-kidger:
Oh, follow-up thought: you mentioned that this doesn't occur in 64-bit mode. So I wonder if this is due to some floating-point nondeterminism somewhere? (Although I didn't think that occurred on the CPU.)

Randl:
There was non-determinism in operator generation, which I fixed by passing `getkey` to the function instead of using a separate one.
As for now, the generated operators appear to be identical.

All of this happens in 64-bit mode but happens only to complex inputs (`complex128`) and not floats (`float64`).

It may be jax bug, because I have no idea where else non-determinism can be.

patrick-kidger:
Right, just to be clear, I'm not talking about nondeterminism in creating the `JacobianLinearOperator` object. I'm hoping we can track the nondeterminism through the internals of the `mv` call -- which primitive operation is actually producing the nondeterminism? We should be able to reduce this to a MWE of calling just this one operation.

Randl:
Right, the `mv` in this case is unlikely to be the problem. Since we call `linearise` in NormalCG
https://github.com/google/lineax/blob/199c27777dea454f6a7d23c531664dba9bd2cbf9/lineax/_solver/cg.py#L115
the operator, in this case, is `FunctionLinearOperator`, for which `mv` is just a function application.
https://github.com/google/lineax/blob/199c27777dea454f6a7d23c531664dba9bd2cbf9/lineax/_operator.py#L634-L635
So I guess the operator is corrupted somewhere on the way (in linearise maybe, which is just `jax.linearize`).
Reducing this to a MWE of calling just this one operation may be impossible -- e.g., adding prints in "wrong" places makes it disappear.

patrick-kidger:
Yeah, I'm afraid complex support isn't super high priority for me right now, so I definitely don't have time to dig into this myself!

What do you mean that fixing the seed makes the bug disappear? My understanding is that all of this was always for a fixed seed.

Randl:
I mean that if I run test as-is, it'll fail ~2% of times. If I then will take the seed from output of one of the failed tests and paste in the getkey instead of random number call
https://github.com/google/lineax/blob/199c27777dea454f6a7d23c531664dba9bd2cbf9/tests/conftest.py#L47-L49
it won't fail (or fails at least 20 times less often). The generated operator seems to be the same and the problematic `mv` result is different to one of the failed test but identical between successful tests

Randl:
Oh, since I run repeated tests inside one run, I missed what happened.
If I put a constant failed seed in `conftest.py`, the test fails in ~25% of runs, but inside the run, it either fails every time or succeeds every time, no matter how many runs.
Adding prints for the operator inside CG still fixes the test.

patrick-kidger:
Right! I think that makes sense -- if this is due to some oddity in the compilation process, then repeated tests inside a single run are probably all hitting the same compilation cache.

I wouldn't worry about trying to debug it with prints etc. -- just look at the final output of the overall program. Then keep chopping pieces out of the program (never mind whether it's still doing a linear solve or not) to simplify it down to the smallest example of a nondeterministic program we can find.

Randl:
I've digged a bit deeper and generated HLO is identical so I suspect the bug is somewhere outside Jax (XLA? I filed a bug report for them https://github.com/openxla/xla/issues/8471)

patrick-kidger:
Ah, I spent some time digging into this as well and found something similar -- I likewise found that the wrong answer was sometimes produced a few percent of the time when running on Ubuntu and on CPU. (Unfortunately I think I've lost access to the computer where I had this repro script now though.)

This is definitely an XLA bug.

Would you be able to reduce your bug report to a copy-pastable MWE? Maybe provide just the HLO as some copy-pastable text. That'll help a lot for the folks there to be able to reproduce it. Something like a whole repo is a much larger thing for them to try and understand.

Randl:
So I don't know how to run HLO to verify that, but I've attached compiled HLO to my bug report in XLA. 
I'm fairly sure it should be reproducible from that file.

Randl:
So as the bug was fixed in https://github.com/google/jax/commit/4a35c12077ca427ed61f8a6b42b41e5f7bcaf876 https://github.com/openxla/xla/pull/10501 (I verified that it is not failing on latest nightly), I think we can try to merge it now.

I had to add some casts due to strict typing. It looks a bit awkward, but I'm not aware of a better way to extract pytree dtype.
Also some tests failed after added casts, which is weird, I'll be looking into it later; if you have any ideas I'd love to hear them.

patrick-kidger:
Yup, I'm really glad that this issue was finally resolved! That was a thorny one.
(And thank you for so carefully bisecting it, that can't have been easy.)

I think we'll have to wait until a new jaxlib release before any of the tests in the CI pass. Let's revisit this then, as that'll be the canonical thing that determines whether or not we can merge this :)

Randl:
Just to clarify, I don't think the current test failures are related to the resolved bug: first, they are reproduced on the latest nightly; second, they are reproduced every time and not stochastically like the bug in question.

patrick-kidger:
Swapping the inner product order makes sense to me.

The dtype casting I'm less sure about. It seems to just use whatever the dtype of the first leaf is, which may not be the largest dtype.

FWIW, right now we do have a handful of tests that we reproduce mixed dtypes correctly:

https://github.com/patrick-kidger/lineax/blob/9b923c8df6556551fedc7adeea7979b5c7b3ffb0/tests/test_solve.py#L51-L123

although notably none of the iterable solvers are tested there at the moment. (I don't remember why, that seems like a glaring omission.) And so in particular I'm wondering whether we should allow for different parts of the state to be different dtypes, rather than simply upcasting everything to a single dtype. WDYT?

Randl:
To clarify, without casts, the binary operation between float and complex fails due to strict typing.

Most of casts are for comparisons: thresholds in CG and machine epsilon; both should be safe to cast to larger type I think and were already float64.

One notable exception is scalar-operator multiplication, there probably using largest type makes sense tho.

As for specific implementation, I would like to use something like [`common_type`](https://numpy.org/doc/stable/reference/generated/numpy.common_type.html) but apparently it doesn't exist in Jax?

patrick-kidger:
I think the analogue for `common_type` would be `jnp.result_type`. (I don't know why NumPy has two essentially-identical functions.)

Randl:
I think we're at the point where we can merge it. Before the Jax release, it is possible that the tests will randomly fail; I think it is easier to rerun the tests in the unlikely case it will happen than maintain this PR without merging.

The next step will be to update the docs/examples as per https://github.com/patrick-kidger/lineax/issues/57; this is also a good point to consider what complex-specific behavior is not tested.
After this and Jax 0.4.26 release, we can release and start tackling the issues down the line, I think, e.g., https://github.com/patrick-kidger/diffrax/pull/391

@patrick-kidger, what do you think?

patrick-kidger:
That all makes sense to me! So on that basis, I've just merged this PR.

Thank you for all your efforts getting this done, it's awesome to see this merged! :)

PR #66, opened 2023-12-04
Jvp and helper fixes

Randl:
Since https://github.com/google/lineax/pull/64 is blocked by a bug, I'm taking functional changes to a separate PR (but not enabling tests for complex variables yet). This includes fixing JVP for `linear_solve`, fixing Cholesky solver transposition, changing the randomness source in helper to be reproducible, and adding `dtype` parametrization to more tests.

### DIFF
--- a/lineax/_solver/cholesky.py
+++ b/lineax/_solver/cholesky.py
@@ -77,8 +77,14 @@ class Cholesky(AbstractLinearSolver[_CholeskyState]):
         return solution, RESULTS.successful, {}
 
     def transpose(self, state: _CholeskyState, options: dict[str, Any]):
-        # Matrix is symmetric anyway
-        return state, options
+        # Matrix is self-adjoint
+        factor, is_nsd = state
+        return (factor.conj(), is_nsd), options
+
+    def conj(self, state: _CholeskyState, options: dict[str, Any]):
+        # Matrix is self-adjoint
+        factor, is_nsd = state
+        return (factor.conj(), is_nsd), options
 
     def allow_dependent_columns(self, operator):
         return False
--- a/lineax/_operator.py
+++ b/lineax/_operator.py
@@ -1988,11 +1988,9 @@ def _(operator):
 
 @conj.register(TangentLinearOperator)
 def _(operator):
-    # Should be unreachable: TangentLinearOperator is used for a narrow set of
-    # operations only (mv; transpose) inside the JVP rule linear_solve_p.
-    raise NotImplementedError(
-        "Please open a GitHub issue: https://github.com/google/lineax"
-    )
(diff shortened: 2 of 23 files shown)


lineax/_solver/cholesky.py
```diff
@@ -77,8 +77,14 @@ def compute(
         return solution, RESULTS.successful, {}
 
     def transpose(self, state: _CholeskyState, options: dict[str, Any]):
-        # Matrix is symmetric anyway
-        return state, options
+        # Matrix is self-adjoint
+        factor, is_nsd = state
+        return (factor.conj(), is_nsd), options
```
patrick-kidger, line 82: This is a good catch!
This isn't the only solver assuming the the matrix is PSD/symmetric/etc -- I think some of the others may also need updating?
(This is kind of finickity isn't it?)

patrick-kidger (commented):
Nitty comments aside I think this LGTM!
I like that this is a way to make forward progress whilst we try to figure out #64.

lineax/_solver/cholesky.py
patrick-kidger: Definitely possible that the tests need enhancing to catch this -- after all we don't really have any complex tests at all, since that's not really a feature yet.
The fact that we only ever do `conj(transpose(...))` as a pair might also cover some mistakes if the pair is correct even when each individual one is wrong.

I think it'd worth adding some complex tests that explicitly check calling these methods individually.

patrick-kidger: Hmm. At the very least I'd expect the CG tests to fail given complex input: this also requires the operator to be positive or negative definite. I think this indicates we need a lot more tests for complex input!

(And, side note, CG currently appears to have a pointless no-op transpose in the `normal=False` case -- although that's an existing thing, not something changed in this PR.)

Randl: I've patched psd matrix generation earlier:
https://github.com/google/lineax/blob/199c27777dea454f6a7d23c531664dba9bd2cbf9/tests/helpers.py#L54-L57
so I don't thing CG should fail.

The problem here was in implicit assumption psd->symmetric

Randl: By briefly checking other places mentioning symmetric operators I think only this one is also problematic:
https://github.com/google/lineax/blob/199c27777dea454f6a7d23c531664dba9bd2cbf9/lineax/_operator.py#L1407-L1416

Randl: Specifically all operator/solver transpositions look ok to me

patrick-kidger:
Alright, this LGTM! Thanks for these fixes, we're now much closer to full complex support.

PR #75, opened 2024-01-09
added flag whether to use forward or reverse AD in jacobian operator

thibmonsel:


lineax/_operator.py
```diff
@@ -538,6 +529,7 @@ def __init__(
         tags: Union[object, Iterable[object]] = (),
         closure_convert: bool = True,
         _has_aux: bool = False,  # TODO(kidger): remove, no longer used
+        use_jacrev: Union[bool, None] = None,
```
patrick-kidger, line 532: I think I'd suggest a name that doesn't mention just `rev` in particular. E.g. something like
```python
jac: Optional[[Literal["fwd", "bwd"]]
```

lineax/_misc.py
```diff
@@ -45,9 +45,15 @@ def __call__(self, *args, **kwargs):
...
-def jacobian(fn, in_size, out_size, holomorphic=False, has_aux=False):
+def jacobian(fn, in_size, out_size, holomorphic=False, has_aux=False, jac=None):
     # Heuristic for which is better in each case
     # These could probably be tuned a lot more.
+    if jac is not None:
+        if jac == "bwd":
+            return jax.jacrev(fn, holomorphic=holomorphic, has_aux=has_aux)
+        if jac == "fwd":
+            return jax.jacfwd(fn, holomorphic=holomorphic, has_aux=has_aux)
+
```
patrick-kidger, line 56: I think this can be made a little more robust to passing invalid arguments:
```python
if jac is None:
    jac_fwd = (in_size < 100) or (in_size <= 1.5 * out_size)
elif jac == "fwd":
    jac_fwd = True
elif jac == "bwd":
    jac_fwd = False
else:
    raise ValueError("`jac` should either be None, 'fwd', or 'bwd'.")
if jac_fwd:
    return jax.jacfwd(...)
else:
    return jax.jacrev(...)
```

thibmonsel:
Hi Patrick, is this PR missing anything or it looks good to you before merging ?

patrick-kidger:
This LGTM! Thank you for adding this in :)

PR #86, opened 2024-03-16
Implementation of LSMR for iterative least squares.

### LINKED ISSUE #85: Iterative least squares solvers
Hi @patrick-kidger , I've really enjoyed playing around with Lineax!

You have several iterative methods for square systems (CG, GMRES, BICGSTAB) but nothing for least squares problems.

I have a JAX port of `scipy.sparse.linalg.lsmr` that I should be able to get working with lineax, but wanted to check first:

- is this something lineax users would be interested in?
- any possible licensing issues with scipy? it's been partially rewritten for different control flow etc, but I'm always a bit paranoid about software copyright stuff
- What sorts of tests are needed? I've tested it for basic array inputs but not pytrees, vmapping, implicit solves etc.

f0uriest:
Tests seem to be passing locally. I added `LSMR` in the places it seemed to make sense (mainly places where I saw `SVD` which solves similar problems. Let me know if there are other tests that should be added.

A few things that could use some input:

- There are a large number of state variables to keep track of. Right now I'm just using nested tuples but there's probably a more elegant way.

Resolves #85


lineax/_solver/lsmr.py
```diff
@@ -0,0 +1,583 @@
...
+    - `y0`: The initial estimate of the solution to the linear system. Defaults to all
+        zeros.
+    - `damp`: Damping factor for regularized least-squares. LSMR solves the regularized
+        least-squares problem:
+
+            min ||A y - b||_2 + damp*||y||_2
+
+        where damp is a scalar. If damp is None or 0, the system is solved without
+        regularization. Default is 0.
+
```
patrick-kidger, line 76: Nit: spurious newline.

lineax/_solver/lsmr.py
```diff
@@ -0,0 +1,583 @@
...
+    """
+
+    atol: float
+    btol: float
```
patrick-kidger, line 80: The convention we've been using through the Equinox scientific ecosystem is to use `atol` (absolute tolerance) and `rtol` (relative tolerance). Different parts of the numerical community keep cooking up different kinds of convergence criteria for different linear/nonlinear/etc. problems, so we're attempting to impose a little order :D

Would it be possible to switch this over in the same way?

FWIW you should be able to arrange for a solver to take exactly `n` steps by setting `atol=rtol=0` and `max_steps=n`, `lx.linear_solve(..., throw=False)`. This makes it possible to compare numerical results across implementations, if that is desired.

lineax/_solver/lsmr.py
```diff
@@ -0,0 +1,583 @@
...
+        vector: PyTree[Array],
+        options: dict[str, Any],
+    ) -> tuple[PyTree[Array], RESULTS, dict[str, Any]]:
+        operator = state
+        x = options.get("y0", None)
+        damp = options.get("damp", 0.0)
+
+        m, n = operator.out_size(), operator.in_size()
+        # number of singular values
+        minDim = min([m, n])
```
patrick-kidger, line 113: Let's use `snake_case` as usual in Python.

lineax/_solver/lsmr.py
```diff
@@ -0,0 +1,583 @@
...
+            v = operator.T.mv(u)
+            alpha = self.norm(v)
+            return u, v, alpha
+
+        def beta_le0(beta, u):
+            v = jtu.tree_map(jnp.zeros_like, operator.in_structure())
+            alpha = 0.0
+            return u, v, alpha
+
+        u, v, alpha = lax.cond(beta > 0, beta_gt0, beta_le0, beta, u)
```
patrick-kidger, line 138: Beta is a norm, how can be it be `<=0`, as then name suggests? (Is this check meant to specifically handle the `=0` case?)

lineax/_solver/lsmr.py
```diff
@@ -0,0 +1,583 @@
...
+        v = lax.cond(alpha > 0, lambda: (ω(v) / alpha).ω, lambda: v)
+
+        h = v.copy()
```
patrick-kidger, line 141: FYI, I don't believe this does anything in JAX, just `h = v` would suffice.

lineax/_solver/lsmr.py
```diff
@@ -0,0 +1,583 @@
...
+        normA = jnp.sqrt(normA2)
+        condA = 1.0
+
+        loop_state_anorm = (normA2, maxrbar, minrbar, normA, condA)
+
+        # Items for use in stopping rules, normb set earlier
+        istop = 0
+        ctol = 0.0
+        if self.conlim > 0:
+            ctol = 1 / self.conlim
```
patrick-kidger, line 181: I'd prefer single-assignment form, i.e.
```python
if self.conlim > 0:
    ctol = 1 / self.conlim
else:
    ctol = 0
```

patrick-kidger: Also perhaps this should be a `jnp.where`, just to support the (very edge) case that `self.conlim` is a traced value?

lineax/_solver/lsmr.py
```diff
@@ -0,0 +1,583 @@
...
+
+        loop_state_stopping = (istop, normr, normb)
+
+        loop_state = (
+            loop_state_main,
+            loop_state_r_est,
+            loop_state_anorm,
+            loop_state_stopping,
+            loop_state_vecs,
+        )
```
patrick-kidger, line 192: Gosh, this is a lot! To answer your question about state variables: for this I would use a module. E.g.
```python
class Foo(eqx.Module):
    bar: ...
    baz: ...
```
which work perfectly well as JAX-compatible containers for state.

lineax/_solver/lsmr.py
```diff
@@ -0,0 +1,583 @@
...
+            test1 = normr / normb
+            test2 = lax.select((normA * normr) != 0, normar / (normA * normr), jnp.inf)
+            test3 = 1 / condA
+            t1 = test1 / (1 + normA * normx / normb)
+            rtol = self.btol + self.atol * normA * normx / normb
+
+            # bitmask for istop for possibly multiple conditions?
+
+            # x is an approximate solution to A@x = b, according to atol and btol.
+            istop += (test1 <= rtol).astype(int) * 2**1
```
patrick-kidger, line 241: Nit: in general, prefer an explicit `jnp....` dtype rather than a plain `int`, as this actually has different precisions between Windows and Unix.
(It doesn't really matter here of course, where the numbers are small.)

lineax/_solver/lsmr.py
```diff
@@ -0,0 +1,583 @@
...
+            istop += (1 + t1 <= 1).astype(int) * 2**4
+            # x approximately solves the least-squares problem according to atol=eps.
+            istop += (1 + test2 <= 1).astype(int) * 2**5
+            # cond(A) seems to be greater than 1/eps
+            istop += (1 + test3 <= 1).astype(int) * 2**6
+
+            # maxiter exceeded
+            istop += (itn >= max_steps).astype(int) * 2**7
+
+            return istop == 0
```
patrick-kidger, line 262: This seems like an odd way to test multiple conditions. I'm guessing this is coming from some older system which didn't support bool types.

I think here we probably just want `return (itn < max_steps) & (1 + test3 > 1) & ...` ?

lineax/_solver/lsmr.py
```diff
@@ -0,0 +1,583 @@
...
+            u = (ω(u) + ω(operator.mv(v))).ω
+            beta = self.norm(u)
+
+            def beta_gt0(alpha, beta, u, v):
+                u = (ω(u) * (1 / beta)).ω
+                v = (ω(v) * -beta).ω
+                Au = operator.T.mv(u)
+                v = (ω(v) + ω(Au)).ω
+                alpha = self.norm(v)
+                v = lax.cond(alpha > 0, lambda x: (ω(x) / alpha).ω, lambda x: x, v)
```
patrick-kidger, line 314: Again I don't really understand this check -- specifically just for the case that `alpha = 0`?

FWIW I think this one has the potential to misbehave. There are circumstances (most notably when vmap'd) under which a `cond` will turn into a `select`, and so both branches are evaluated. In particular that means we will be doing a division-by-zero here, which is probably worth avoiding.

I'd probably suggest a double-where trick:
```
alpha = ...
pred = alpha == 0
safe_alpha = jnp.where(pred, 1 , alpha)
v = jtu.tree_map(lambda vi: jnp.where(pred, 0, vi / safe_alpha), v)
```

lineax/_solver/lsmr.py
```diff
@@ -0,0 +1,583 @@
...
+        """
+
+        def bzero(a, b):
+            return jnp.sign(a), 0.0, abs(a)
+
+        def azero(a, b):
+            return 0.0, jnp.sign(b), abs(b)
+
+        def b_gt_a(a, b):
+            tau = a / b
```
patrick-kidger, line 529: I suspect this could do something funny for the same reason as above -- we may actually have `a = 0` here.
I think whenever dividing by `a` or `b` it should probably be replaced with something like `/ jnp.where(b==0, 1, b)` (and then the outer `cond` will ensure that the `b=0` case is still handled correctly, by just ignoring the results of this computation).

lineax/_solver/lsmr.py
```diff
@@ -0,0 +1,583 @@
...
+
+- `atol`: Relative tolerance (relative to norm(A*x)) for terminating solve.
+- `btol`: Relative tolerance (relative to norm(b)) for terminating solve.
+- `norm`: The norm to use when computing whether the error falls within the tolerance.
+    Defaults to the two norm.
+- `max_steps`: The maximum number of iterations to run the solver for. If more steps
+    than this are required, then the solve is halted with a failure.
+- `conlim`: The solver terminates if an estimate of cond(A) exceeds conlim. For
+    compatible systems Ax = b, conlim could be as large as 1.0e+12 (say). For
+    least-squares problems, conlim should be less than 1.0e+8. If conlim is None,
```
patrick-kidger, line 579: I don't think `None` is handled?

patrick-kidger (commented):
This looks really cool! Thank you for contributing it, I can see this must have been a lot of effort. :D I've gone through and left various comments, which should all be very small. In addition to those, can we add `LSMR` to the well-posed square tests?

On the exit conditions: how many does LSMR potentially have?

On complex dtypes: indeed, but I think we now have things almost entirely working. Should probably get that fixed on our end in the next few couple of weeks, and then we're just waiting on a new JAX release so that the upstream fixes there are generally available. (I think we should probably let all of that happen before merging this PR, so that we can add LSMR to the complex tests as well.)

lineax/_solver/lsmr.py
f0uriest: The main reason for this was so that after exiting the loop we actually know which condition caused the exit, since there are in theory a few different options. Another option would be to add specific flags to the state that get set but the state is already pretty huge...

lineax/_solver/lsmr.py
f0uriest: I had tried that, but ran into the issue that equinox modules are immutable, so simple operations like

```python
state.alpha = state.beta * state.x
```

need a lot of pytree surgery. Do you know of any tricks that could help?

lineax/_solver/lsmr.py
f0uriest: yeah this is really checking the beta==0 case, I'll rename things to make it clearer

lineax/_solver/lsmr.py
f0uriest: These are a bit confusingly named, since they are both actually relative tolerances, but relative to `a` (the linear operator) or `b` (the rhs vector). I think we could probably change it, since later on they both get reduced to a single relative tolerance:

```python
            rtol = self.btol + self.atol * normA * normx / normb
```

I'd have to think about an absolute tolerance, since it's a bit weird for least squares where the residual is never expected to be zero.

f0uriest:
The original version has 7 possible exit conditions:

1. exceeds maxiter
2. x solve Ax=b to user specified tolerance (ie, no least squares needed, problem appears well posed)
3. x is the least squares solution to user specified tolerance
4. cond(A) larger than user specified conlim
5. x solve Ax=b to tolerance=eps (same as 2, but guards against users setting tolerances to zero?)
6. x is the least squares solution to tolerance=eps (same as 3?)
7. cond(A) > 1/eps (same as 4?)

I think we could get rid of 5-7 provided that maxiter is always finite. 2 and 3 could probably be combined into a standard "successful" exit assuming the user doesn't really care whether its the "true" solution or least squares. The main new one would be the condition number one. We could maybe shoehorn that into `lx.RESULTS.singular` or `lx.RESULTS.breakdown`? or could add a new one just for that.

andycasey:
I know it's bad form to comment just when you want to 👍🏻 , but having LSMR basically made my problem go from "untractable and 10x slower with jax than scipy" to tractable. 

The problem here is one where I have a large linear system and I never want to construct the design matrix, so iterative methods are all I can really use. QR or SVD had huge memory overheads and were very slow. The Jax implementation for GMRES (in `lineax` or `jax.scipy.sparse.linalg.gmres`) doesn't seem to accept non-square matrices, although the `scipy.sparse.linalg.gmres` implementation does.

So this is a big 👍🏻 to this feature.

lineax/_solver/lsmr.py
f0uriest: After trying a few different options I just went with a dict. Happy to change it if needed.

PTNobel:
Would it be possible to pick up this PR? Are there outstanding TODOs? Otherwise, I'm happy to start by rebasing it.

lineax/_solver/lsmr.py
```diff
@@ -0,0 +1,440 @@
...
+from .._norm import two_norm
+from .._operator import AbstractLinearOperator, conj
+from .._solution import RESULTS
+from .._solve import AbstractLinearSolver
+
+
+_LSMRState: TypeAlias = AbstractLinearOperator
+
+
+class LSMR(AbstractLinearSolver[_LSMRState], strict=True):
```
PTNobel, line 55: I think `strict=True` is no longer used in lineax; see #155 

Overall, `strict` just opts into extra error validation and so shouldn't change the runtime behavior.

PTNobel (commented):
Did a pass through the code (tho I skipped the actual math of the implementation) and it looks good to me.

I'm open to doing a review of the implementation itself if that would help the PR move along. I'd probably verify the implementation follows the scipy implementation rather than compare it to the original paper if that would be acceptable.

patrick-kidger:
This is totally on me, this has languished at the bottom of my GitHub to-do list for a long time. To this point:

> I'm open to doing a review of the implementation itself if that would help the PR move along. I'd probably verify the implementation follows the scipy implementation rather than compare it to the original paper if that would be acceptable.

This is absolutely the reason it's taken so long to get this in, as I've never found the time to sit down and learn the algorithim + understand why it works + make sure that this conforms to it.

Over time we've had much more of a community form around these projects, however, and I think I'm learning to better delegate this kind of verification to the community. 😄 If you'd be interested in taking this on then that would definitely help get this in. (Is there a test we can write to ensure this / to ensure it stays like this? We already have test for performing the linsolve itself so maybe there's not much to be added there.)

In terms of code quality I have no notes, the implementation looks excellent.

healeyq3:
> Maybe more? @healeyq3 can you put together a good list of test cases to implement?

@PTNobel will do! I'm at work right now, but will aim to have done by EOD PST. Worst case will finish tomorrow.

lineax/_solver/lsmr.py
```diff
@@ -0,0 +1,440 @@
...
+from equinox.internal import ω
+from jaxtyping import Array, PyTree
+
+from .._norm import two_norm
+from .._operator import AbstractLinearOperator, conj
+from .._solution import RESULTS
+from .._solve import AbstractLinearSolver
+
+
+_LSMRState: TypeAlias = AbstractLinearOperator
```
PTNobel, line 52:
```suggestion
type _LSMRState = AbstractLinearOperator
```

healeyq3:
> > Maybe more? @healeyq3 can you put together a good list of test cases to implement?
> 
> @PTNobel will do! I'm at work right now, but will aim to have done by EOD PST. Worst case will finish tomorrow.

I’ll need to continue to work this tomorrow, but @PTNobel here are my current notes on testing.

With the code changes in this PR, the following will be tested:
1. (added explicitly)`test_iterative_solver_max_steps_only`, which is not testing any LSMR accuracy. 
2. (added explicitly) `test_nonsquare_pytree_operator1 ` and `test_nonsquare_pytree_operator2 `, which are just testing LSMR vs JAX `lstsq` output on small **singular** matrices (embedded in PyTrees).
3. (added via addition of LSMR to `solvers`/`params`) `test_small_wellposed`, `test_pytree_wellposed`, and `test_small_singular`

You've proposed that we add tests for the following:
1. $Ax = b$ where $b \not \in \text{Col}(A)$, and that it correctly halts iterations when the least squares solution is found.
2. $A$ tall
3. $A$ wide
4. $A$ with dependent columns, such that the least-norm least-squares is well-defined and verify it generates that
5. Tests with $\text{damp} = 0$.

Some other tests we may want to consider:
1. Solving a larger, well-posed system
2. Monotonically decreasing $\|A^T r_k \|$, where $r_k$ is the $k\text{th}$ residual (and $\|\cdot\|$ denotes a norm...I can't figure out why that isn't rendering properly)
3. Check if the regularized LSMR has better convergence than the non-regularized LSMR on an ill-conditioned A
4. Explicitly test all stopping criteria (residual norm, condition estimate, maximum iterations; although the 1st and 3rd should be covered by other tests already)

Let me know what you think--I can return to this tomorrow night (I have obligations during the day, otherwise would do sooner. I'm also entirely free this weekend, so happy to help implement these.

lineax/_solver/lsmr.py
```diff
@@ -0,0 +1,440 @@
...
+            and isinstance(self.rtol, (int, float))
+            and self.atol == 0
+            and self.rtol == 0
+        )
+
+        m, n = operator.out_size(), operator.in_size()
+        # number of singular values
+        min_dim = min([m, n])
+        if self.max_steps is None:
+            max_steps = min_dim * 10  # for consistency with other iterative solvers
```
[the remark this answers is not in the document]
PTNobel: I did some testing on some random matrices and the SCIPY maxiter count actually caused the problem to not converge when it would have if we'd even done 2 * min_dim. So I'm sold we should be consistent with the other Krylov methods and not scipy.

lineax/_solver/lsmr.py
```diff
@@ -0,0 +1,440 @@
...
+
+    This supports the following `options` (as passed to
+    `lx.linear_solve(..., options=...)`).
+
+    - `y0`: The initial estimate of the solution to the linear system. Defaults to all
+        zeros.
+    - `damp`: Damping factor for regularized least-squares. LSMR solves the regularized
+        least-squares problem:
+
+            min ||A y - b||_2 + damp*||y||_2
```
PTNobel, line 71: See my PR but these docs are wrong. Here's the easiest fix:
```suggestion
            min ||A y - b||_2^2 + damp*||y||_2^2
```

patrick-kidger:
> I'm happy to add the test if you think it is necessary to merge this, but I don't think it is worth doing. It tests properties of the LSMR iteration that are almost certainly correct if the implementation is actually working. I generally try to verify that iterative algorithms work by seeing how many iterations result in correct solutions, rather than to test the iterates themselves. Complete respect if you disagree tho.

No opinion from me, happy to defer to @healeyq3 on this one.

(Just noting how it can be done if it should be done.)

healeyq3:
> No opinion from me, happy to defer to @healeyq3 on this one.

What has already been agreed upon is a (IMO) rather comprehensive test suite, so certainly don't add this test on my account! This can be used to debug (?) if need be, but I doubt we come to that.

lineax/_solver/lsmr.py
```diff
@@ -0,0 +1,440 @@
...
+            *jtu.tree_leaves(operator.in_structure()),
+        )
+
+        b = vector
+        u = (ω(b) - ω(operator.mv(x))).ω
+        normb = self.norm(b)
+        beta = self.norm(u)
+
+        def beta_nonzero(beta, u):
+            u = (ω(u) / lax.select(beta == 0.0, 1.0, beta).astype(dtype)).ω
```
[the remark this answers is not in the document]
f0uriest: This is basically a version of the [double where trick](https://docs.jax.dev/en/latest/faq.html#gradients-contain-nan-where-using-where) to avoid nans in reverse mode AD when dealing with division by zero.

It's not strictly necessary here because in `lineax` we never differentiate through the solver itself, but I still think it's good practice in case we ever add some sort of "direct adjoint" or if someone is using this outside of `lineax.linear_solve`

lineax/_solver/lsmr.py
```diff
@@ -0,0 +1,440 @@
...
+        stats = {
+            "num_steps": loop_state["itn"],
+            "istop": loop_state["istop"],
+            "norm_r": loop_state["normr"],
+            "norm_Ar": loop_state["normAr"],
+            "norm_A": loop_state["normA"],
+            "cond_A": loop_state["condA"],
+            "norm_x": self.norm(loop_state["x"]),
+        }
+        if (self.max_steps is None) or (max_steps < self.max_steps):
```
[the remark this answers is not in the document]
f0uriest: This was mostly copied from the other iterative solvers, @patrick-kidger might be able to say more about the different conditions here

lineax/_solver/lsmr.py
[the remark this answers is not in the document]
PTNobel: Ah that makes sense! Thanks for the reference.

lineax/_solver/lsmr.py
```diff
@@ -0,0 +1,438 @@
...
+
+    def compute(
+        self,
+        state: _LSMRState,
+        vector: PyTree[Array],
+        options: dict[str, Any],
+    ) -> tuple[PyTree[Array], RESULTS, dict[str, Any]]:
+        operator = state
+        x = options.get("y0", None)
+        damp = options.get("damp", 0.0)
```
PTNobel, line 110:
```suggestion
        damp = options.get("damp", 0.0)
        assert damp == 0.0, "Damp support is not implemented yet" 
```

PTNobel:
Thinking about this more; I think we should consider disabling damp support and just merging. It isn't clear to me how we're supposed to implement transpose when damp != 0, and I think this PR is ready to go without that feature.

@patrick-kidger would you be open to merging without damp support?

PTNobel:
@patrick-kidger I think this is good to merge. The CI failures look spurious/irrelevant to me.

johannahaffner:
I think a segfault matters :) can take a look this afternoon, might be similar to one that I chased a while back.

PR #87, opened 2024-03-18
Fixed and added tests for JacobianLinearOperator(jac="bwd")

patrick-kidger:


lineax/_operator.py
```diff
@@ -568,7 +568,14 @@ def __init__(
 
     def mv(self, vector):
         fn = _NoAuxOut(_NoAuxIn(self.fn, self.args))
-        _, out = jax.jvp(fn, (self.x,), (vector,))
+        if self.jac == "fwd":
+            _, out = jax.jvp(fn, (self.x,), (vector,))
+        else:
+            assert self.jac == "bwd"
```
Randl, line 574: Wouldn't it be better to do something like
```python
       if self.jac == "fwd":
            ...
       elif self.jac == "bwd":
            ...
       else:
            raise ValueError('jac should be "fwd" or "bwd"')
```
?

patrick-kidger: Agreed! Just done.

PR #89, opened 2024-03-22
Add casts in additional place to satisfy strict typing for complex case

Randl:
Split off https://github.com/patrick-kidger/lineax/pull/64, see https://github.com/patrick-kidger/lineax/issues/57

### DIFF
--- a/lineax/_misc.py
+++ b/lineax/_misc.py
@@ -21,8 +21,6 @@ import jax.numpy as jnp
 import jax.tree_util as jtu
 from jaxtyping import Array, ArrayLike, Bool, PyTree  # pyright:ignore
 
-from ._norm import default_floating_dtype
-
 
 def tree_where(
     pred: Bool[ArrayLike, ""], true: PyTree[ArrayLike], false: PyTree[ArrayLike]
@@ -78,6 +76,8 @@ def _asarray_jvp(dtype, x, tx):
 
 
 def inexact_asarray(x):
+    from ._norm import default_floating_dtype
+
     dtype = jnp.result_type(x)
     if not jnp.issubdtype(jnp.result_type(x), jnp.inexact):
         dtype = default_floating_dtype()
--- a/lineax/_misc.py
+++ b/lineax/_misc.py
@@ -82,3 +82,7 @@ def inexact_asarray(x):
     if not jnp.issubdtype(jnp.result_type(x), jnp.inexact):
         dtype = default_floating_dtype()
     return _asarray(dtype, x)
+
+
+def complex_to_real_dtype(dtype):
+    return jnp.finfo(dtype).dtype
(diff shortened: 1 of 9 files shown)


Randl:
Looks like that the failed test is just unlucky seed?

patrick-kidger:
So I think it's probably best to avoid casting absolutely everything to the same dtype, as discussed over in #64.

Probably the simplest thing to do is just to locally allow dtype casting -- I do this in a few places where it's awkward not to:
```python
with jax.numpy_dtype_promotion("standard"):
    ...
```

As for an unlucky seed -- yup, that seems plausible. GMRES is suck a finickity algorithm. I'm not sure I have a nice way to guard against this error, so I'm happy to just ignore it.

Randl:
So there are four cases where this PR introduces casts:
1. Finite difference in test helpers. 
2. Scalar in `DivLinearOperator`operator
3. Scales in CG and friends
4. eps in GMRES normalization.

So I think for 2 and 4 it makes sense to cast since the value multiplies the input, and for 1 it doesn't really matter. No strong opinion about 3. What do you think?

patrick-kidger:
I think I'd probably suggest using `with jax.numpy_dtype_promotion("standard")` for all of them, actually! For example on 2: note that right now, if you have a pytree of `(f32, c64)` then the scalar will get promoted to `result_type = c64` and as such the entire output will be as well.

lineax/_solver/gmres.py
```diff
@@ -395,13 +400,17 @@ def _arnoldi_gram_schmidt(
...
     def _normalise(
         self, x: PyTree[Array], eps: Optional[Float[ArrayLike, ""]]
-    ) -> tuple[PyTree[Array], Float[Array, ""], Bool[ArrayLike, ""]]:
+    ) -> tuple[PyTree[Array], Inexact[Array, ""], Bool[ArrayLike, ""]]:
+        input_dtype = jnp.result_type(*jtu.tree_leaves(x))
         norm = two_norm(x)
         if eps is None:
             eps = jnp.finfo(norm.dtype).eps
+        else:
+            eps = jnp.astype(eps, jnp.finfo(input_dtype).dtype)
```
patrick-kidger, line 409: I think this should probably just use `norm.dtype`?

Randl: Yes

lineax/_misc.py
```diff
@@ -78,7 +76,13 @@ def _asarray_jvp(dtype, x, tx):
 
 
 def inexact_asarray(x):
+    from ._norm import default_floating_dtype
```
patrick-kidger, line 79: I think let's just move `default_floating_dtype` into `_misc.py` rather than have a circular import.

patrick-kidger:
LGTM -- thank you as always! Where are we at now for complex support in Lineax?

PR #91, opened 2024-03-22
Norm should return float rather than complex

Randl:


patrick-kidger:
Hah! This is a good trick.

PR #92, opened 2024-03-31
Add complex-related stuff to docs and examples

Randl:
See #57

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -1,10 +1,10 @@
 # FAQ
 
-## How does this differ to `jax.numpy.solve`, `jax.scipy.{...}` etc.?
+## How does this differ from `jax.numpy.solve`, `jax.scipy.{...}` etc.?
 
 Lineax offers several improvements. Most notably:
 
-- Several new solvers. For example, [`lineax.QR`][] has no counterpart in core JAX. (And is much faster than `jax.numpy.linalg.lstsq`, which is the closest equivalent, and uses an SVD decomposition instead.)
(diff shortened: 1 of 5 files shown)


docs/examples/structured_matrices.ipynb
patrick-kidger, line 1: You seem to have leaked details about your local environment in the error messages here! I try to tidy these up manually if necessary, to avoid that.

FWIW I think probably the best way to advertise our new complex support would be to leave the current examples alone, and instead add a whole-new example dedicated to complex support!

docs/faq.md
patrick-kidger, line 1: Haha, great job fixing up my many typos :D Thank you!

docs/examples/structured_matrices.ipynb
Randl: Fair enough, I'll rework it

Randl:
I've added a separate simple example with complex entries; I can adda more sophisticated one if you have something in mind

patrick-kidger (commented):
This LGTM!

Can you add your extra example into `mkdocs.yml`? Otherwise it won't appear in the generated documentation :)

patrick-kidger:
Alright, wonderful stuff! This LGTM.

As merging to main will update the docs then I'm going to hold off on merging this PR until we're ready to do the next release, but morally speaking this PR should be considered merged! :)
(I have one other upstream change that I want to land in Equinox first, and then I'm hoping to do a new release of both Equinox and Lineax -- so probably the next week or two.)

Issue #96, opened 2024-05-25
Block operators / solvers


f0uriest:
Following the discussion in #80 I thought it would be good to lay out some ideas for a general block matrix/operator abstraction.

Some useful types of block operators:
- Block tridiagonal where each sub-block is dense (#80)
- Block triangular where each sub-block is dense (can be a factor of num_blocks faster than direct inversion)
- Block dense where each sub-block is either dense or identically zero - can use pivoted block LU?

I think the more powerful capability is if we allow block operators where each block itself can be an arbitrary `LinearOperator` including nested block operators. This would allow for a pretty expressive API that could represent most structured sparsity patterns that commonly arise when discretizing PDEs and lots of other applications.

Doing the latter would likely require a few things:
- some sort of symbolic zero operator for representing empty blocks.
- Some sort of `InverseLinearOperator` abstraction, that wraps another linear operator such that `InverseLinearOperator(operator).mv(x) == linear_solve(operator, x)`
- Allow `linear_solve` to accept an operator as the RHS, such that `linear_solve(operator1, operator2) -> ComposedLinearOperator(InverseLinearOperator(operator1), operator2)`
These would allow us to symbolically/algorithmically invert an arbitrary block operator without having to know the specifics of what each block is.

Some other thoughts:
- If each sub-block is dense the overall block operator can usually be stored as 
a dense array with some special indexing to avoid having to store blocks of zeros. This is nice because arrays can be scanned/vmapped etc.
- However, if we allow sub-blocks to be `LinearOperators` than we might have to resort to lists of blocks, which (generally?) can't be scanned/vmapped as easily, so compile times might be longer? This might not be too bad if we generally assume that block_size >> num_blocks.
- For blocks that share a particular structure, can we vmap block creation, instead of having to create each one separately and store them in a list?
- For a given sub-operator, how to decide which solver to use to invert it? Likely this can be done with proper tags using `AutoLinearSolver` but we may want to add some new tags to indicate that `AutoLinearSolver` use an iterative method in some cases etc? Or maybe have `solver` similarly be a blocked/pytree type thing allowing the user to specify which solver to use at each level/block of the operator

patrick-kidger:
Thanks for opening this issue -- I think you've hit pretty much all the major points I can think of, and also touched on several I hadn't, like vmap'ing blocks!

Some initial comments:

[...]

PR #97, opened 2024-05-28
New solver based on Woodbury matrix identity

aidancrilly:
In issue #3 , a solver based on Woodbury matrix identity was given as a new solver feature for lineax.

Here I have made a first attempt at implementing it. It works slightly differently than the other base LinearOperators as it takes a LinearOperator as an argument (for A), with U, C and V as JAX arrays. This allows for the correct specialised solvers to be used for inverse(A) operations. E.g. A can be a DiagonalLinearOperator/TridiagonalLinearOperator/etc. - see test_Woodbury in test_operator.py

Things I am not certain of:
1. Is passing solvers and states around in the SolverState of Woodbury the best approach. It would seem to me to minimise re-initialising solvers for multiple RHS. However perhaps there is a neater solution. 
2. I have not included a tag for Woodbury, it seems difficult to identify if matrices have this structure so it seems to me it should be left to the user. However, I am not sure if I have handled the operation-type linear operators (Add,Mul,Div,etc.) correctly in this instance.
3. I had to add a pyright ignore in _solve.py under _linear_solve_transpose for allow_struct, not sure why.
```
ft.partial(eqxi.materialise_zeros, allow_struct=True), # pyright: ignore
```
But it made me notice that materialise_zeros is from equinox 0.11.4 but pyproject.toml requires >=0.11.3. Does this require correcting?


lineax/_operator.py
```diff
@@ -906,6 +906,78 @@ def out_structure(self):
...
+    UCV: Array
+    tags: frozenset[object] = eqx.field(static=True)
+
+    def __init__(
+        self,
+        A: AbstractLinearOperator,
+        C: Inexact[Array, " k k"],
+        U: Inexact[Array, " n k"],
+        V: Inexact[Array, " k n"],
+        tags: Union[object, frozenset[object]] = (),
```
patrick-kidger, line 926: I think we probably shouldn't have any tags at all here.

The tags are our kind of ad-hoc way of avoiding needing to create every combination of `{Diagonal,PositiveDefinite,...}{Matrix,PyTree,...}LinearOperator`. Instead we declare as much of the structure as we can via tags.

However in this case we only really have structure that's interesting: the Woodbury structure. We don't really need to combine this with anything else.

lineax/_operator.py
```diff
@@ -906,6 +906,78 @@ def out_structure(self):
...
+        - `U`: A rank-two JAX array. Shape (n,k)
+        - `V`: A rank-two JAX array. Shape (k,n)
+
+        """
+        self.A = A
+        self.C = inexact_asarray(C)
+        self.U = inexact_asarray(U)
+        self.V = inexact_asarray(V)
+        (N, M) = self.A.in_structure(), self.A.out_structure()
+        if N != M:
```
patrick-kidger, line 944: Edge-case: the structures are arbitrary PyTrees, and we don't know whether they might have defined an `__eq__` method that does something else.

Instead we should use `eqx.tree_equal` here.

lineax/_operator.py
```diff
@@ -906,6 +906,78 @@ def out_structure(self):
...
+            raise ValueError(f"expecting square operator for A, got {N} by {M}")
+        (K, L) = self.C.shape
+        if K != L:
+            raise ValueError(f"expecting square operator for C, got {K} by {L}")
+        N = N.shape[0]
```
patrick-kidger, line 949: `N` is an arbitrary PyTree, it doesn't necessarily have a `.shape` attribute.

lineax/_operator.py
```diff
@@ -906,6 +906,78 @@ def out_structure(self):
...
+        if self.U.shape != (N, K):
+            raise ValueError("U does not have consistent shape with A and C")
+        if self.V.shape != (K, N):
+            raise ValueError("V does not have consistent shape with A and C")
+        self.UCV = self.U @ (self.C @ self.V)
```
patrick-kidger, line 954: Storing things like this is probably a bug. If we modify this operator in any other way than through `__init__` (e.g. evolving it under gradient descent) then this invariant may cease to hold.

Better to compute it on-the-fly when required instead.

patrick-kidger: (In addition this is quite space inefficient!)

lineax/_operator.py
```diff
@@ -1429,6 +1505,33 @@ def _(operator):
...
+    operator. Returns one linear operators and three matrices.
+    **Arguments:**
+    - `operator`: a linear operator.
+    **Returns:**
+    A 4-tuple, consisting of
+    - A which is a linear operator
+    - C, U and V which are matrices
+    For all but the Woodbury operator this extraction is not possible
+    """
+    _default_not_implemented("woodbury", operator)
```
patrick-kidger, line 1527: I don't think we actually need this function!

For a `WoodburyLinearOperator` then a user can just access `.{A, C, U, V}` directly. For anything else then they can't extract these components at all.

lineax/_operator.py
```diff
@@ -1557,6 +1663,36 @@ def _(operator):
...
+def is_Woodbury(operator: AbstractLinearOperator) -> bool:
+    """Returns whether an operator is marked as Woodbury.
+    See [the documentation on linear operator tags](../api/tags.md) for more
+    information.
+    **Arguments:**
+    - `operator`: a linear operator.
+    **Returns:**
+    Either `True` or `False.`
+    """
+    _default_not_implemented("is_Woodbury", operator)
```
patrick-kidger, line 1676: Likewise I don't think this function is needed. Just an `isinstance(x, WoodburyLinearOperator` suffices!

(This kind of compile-time dispatch based on type is a super common design pattern in Equinox-and-friends. See [here](https://docs.kidger.site/equinox/pattern/) for a few other notes on this kind of thing.)

lineax/_solver/Woodbury.py
patrick-kidger, line 1: Nit: I know it's a name but the convention would be to still lower case as `woodbury.py` here.

lineax/_solver/Woodbury.py
```diff
@@ -0,0 +1,138 @@
...
+from .misc import (
+    pack_structures,
+    PackedStructures,
+    ravel_vector,
+    transpose_packed_structures,
+    unravel_solution,
+)
+
+
+_Woodbury_State: TypeAlias = tuple[
```
patrick-kidger, line 25: Also on naming conventions: this would normally be `_WoodburyState`, to avoid mixing `CamelCase` and `snake_case`.

lineax/_solver/Woodbury.py
```diff
@@ -0,0 +1,138 @@
...
+    tuple[Array, Array, Array],
+    tuple[AbstractLinearSolver, Any, AbstractLinearSolver, Any],
+    PackedStructures,
+]
+
+
+def compute_pushthrough(A_solver, A_state, C, U, V):
```
patrick-kidger, line 32: Meanwhile this would have a leading underscore to denote the fact that it is private to this file :)

(Also, can you add type annotations here?)

lineax/_solver/Woodbury.py
```diff
@@ -0,0 +1,138 @@
...
+    # Push through ( C^-1 + V A^-1 U) y = x
+    vmapped_solve = jax.vmap(
+        lambda x_vec: A_solver.compute(A_state, x_vec, {})[0], in_axes=1, out_axes=1
+    )
+    pushthrough_mat = jnp.linalg.inv(C) + V @ vmapped_solve(U)
```
patrick-kidger, line 37: Should we make `C` an `AbstractLinearOperator` as well?

lineax/_solver/Woodbury.py
```diff
@@ -0,0 +1,138 @@
...
+        if not is_Woodbury(operator):
+            raise ValueError(
+                "`Woodbury` may only be used for linear solves with A + U C V structure"
+            )
+        else:
+            A, C, U, V = woodbury(operator)
+            if A.in_size() != A.out_size():
+                raise ValueError("""A must be square""")
+            # Find correct solvers and init for A
+            A_solver = AutoLinearSolver(well_posed=True).select_solver(A)
```
patrick-kidger, line 58: I think the choice of solver here should be an argument to `Woodbury` (the solver), defaulting to `AutoLinearSolver`.

And then at that point I don't think manually calling `select_solver` is required (or even defined in general) -- just `self.a_solver.init` should suffice.

patrick-kidger (commented):
Awesome, thank you for putting this together! I really like the look of this.

I've just left an initial review -- let me know what you think :)

lineax/_solver/Woodbury.py
aidancrilly: I originally had C as an AbstractLinearOperator but it always had to be as_matrix() so its inverse can be computed by jnp.linalg. Therefore, it made the most sense that it is given as a matrix at input so that as_matrix operation is not hidden. If some linearoperators could have an inverse method then perhaps it would make sense for C to be an operator?

lineax/_operator.py
aidancrilly: Ok, This is mostly my ignorance of how lineax is using PyTrees as linear operators.

I am not sure how allowing arbitrary PyTrees for A meshes with the requirement of a Woodbury structure. In the context of the Woodbury matrix identity, I would think A would have to have a matrix representation of an n by n matrix. For a PyTree representation then would C, U and V also need to be PyTrees such that each leaf of the tree can be made to have Woodbury structure?

lineax/_solver/Woodbury.py
patrick-kidger: Ah, right!

Yes, you're completely correct. Indeed I think the appropriate thing to do would be to have a built-in notion of the inverse of an operator. (Tagging also @f0uriest as we were discussing this in #96.)

The inverse of an operator is another operator, and it is basically defined by the action of a linear solve. I think that should mean:
```python
class InverseLinearOperator(AbstractLinearOperator):
    operator: AbstractLinearOperator
    solver: AbstractLinearSolver = AutoLinearSolver(well_posed=True)

    def mv(self, vector):
        return linear_solve(self.operator, vector)

   ...  # other methods here
```

lineax/_operator.py
aidancrilly: Just thought I'd response on this since it has been a bit - thanks for the detailed response, it makes sense, I am (slowly) working on changes that would make the Woodbury implementation pytree compatible. 

I think the size checking is easy enough if I have understood you correctly here (PyTree-of-arrays is basically isomorphic to a vector (flatten every array and concatenate them all together). When checking U and V we need to use the in_size of A and C for N and K.

There will need to be a few tree_unflatten's to move between the flattened space (where U and V live) and the pytree input space (where A and C potentially live). This makes the pushthrough operator a bit tricky but should be do-able with a little time.

I suppose my question would be, is there a nice way to wrap this interlink between flattened vector space and pytree space so that implementing this kind of thing will be easier in the future? Does it already exist somewhere outside (or potentially inside) lineax?

lineax/_solver/Woodbury.py
aidancrilly: Ok, this seems fair. I tried to implement this way and just wanted to raise a two points which you might have comments on.

1. The InverseLinearOperator has to go in _solve.py (or at least not _operator.py) to avoid circular imports. This is potentially a little confusing but maybe it is really the right place for it.
2. Having no conjugate method for LinearOperators makes for a bit of head-ache for this one. conj is attached to the solver so you can update the state of the solver for a conjugate inside InverseLinearOperator (similarly for transpose). However, when you do it this way your operator is not updated. The most obvious way to me to implement the as_matrix method was to use jnp.linalg.inv(self.operator.as_matrix()) - but this ignores the solver_state which might have been transposed or conjugated. I suppose my question is, should LinearOperators have a conjugate method or should LinearSolvers have an as_matrix method for the inverse?

This probably should be its own issue, PR, etc.

tests/test_operator.py
```diff
@@ -141,6 +141,71 @@ def test_diagonal(dtype, getkey):
...
+    C = jr.normal(getkey(), (k, k), dtype=dtype)
+    U = jr.normal(getkey(), (N, k), dtype=dtype)
+    V = jr.normal(getkey(), (k, N), dtype=dtype)
+
+    # Full matrix A
+    WB = lx.WoodburyLinearOperator(lx.MatrixLinearOperator(A), C, U, V)
+
+    full_matrix = WB.as_matrix()
+
+    true_x = jr.normal(getkey(), (N,))
```
Randl, line 160:
```suggestion
    true_x = jr.normal(getkey(), (N,), dtype=dtype)
```
and in two other places too, and then you can add `complex128` to type parametrization

PR #103, opened 2024-08-08
Make identity operator return matrix with matching dtype

Randl:


patrick-kidger:
This LGTM! Thank you for the fix.

PR #105, opened 2024-08-17
Follow conjugation convention in `tree_dot`

Randl:
It may break something downstream unfortunately, so updating dependencies should be done carefully

### DIFF
--- a/lineax/_norm.py
+++ b/lineax/_norm.py
@@ -36,8 +36,8 @@ def tree_dot(tree1: PyTree[ArrayLike], tree2: PyTree[ArrayLike]) -> Inexact[Arra
     for leaf1, leaf2 in zip(leaves1, leaves2):
         dots.append(
             jnp.dot(
-                jnp.reshape(leaf1, -1),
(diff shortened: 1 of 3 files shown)


patrick-kidger:
Modulo failing tests -- not sure what's going on with those (?) -- this LGTM.
Once the tests are fixed we can check the Optimistix and Diffrax tests on this PR before merging it, and see if anything looks awry.

patrick-kidger:
Yes, I think we should.

I can see that tests now pass -- are you happy with this PR / should I merge it?

Randl:
I think I am.
Though one possibility to consider is to not do conjugation in `tree_dot` at all but rater leave it to caller, similarly to `jnp.dot`. This may clutter downstream code but will be consistent with the Jax/numpy semantics.

patrick-kidger:
Merged, then!
And that's an interesting idea. In particular given the difficulties we faced in https://github.com/patrick-kidger/optimistix/pull/61, I can see that that might simplify things.

The decision is yours -- I'd be happy to make that change if you think that is what is best.

PR #106, opened 2024-08-17
Interpret scalar transformations as diagonal

NeilGirdhar:


lineax/_operator.py
```diff
@@ -1514,7 +1514,11 @@ def is_diagonal(operator: AbstractLinearOperator) -> bool:
 @is_diagonal.register(JacobianLinearOperator)
 @is_diagonal.register(FunctionLinearOperator)
 def _(operator):
-    return diagonal_tag in operator.tags
+    return (
+        diagonal_tag in operator.tags
+        or operator.in_size() == 1
+        and operator.out_size() == 1
```
patrick-kidger, line 1520: I think it'd be good practice to group these in brackets: `diagonal_tag in operator.tags or (operator.in_size() == 1 and operator.out_size() == 1)`

I don't have the precedence rules for `x or y and z` in my head!

patrick-kidger (commented):
This LGTM! Thank you for getting the tridiagonal one too, I hadn't noticed that a tridiagonal operator with length-zero off-diagonals should also be possible.

Would you be able to add some tests for this edge case?

lineax/_operator.py
NeilGirdhar: Done!

NeilGirdhar:
> Would you be able to add some tests for this edge case?

Done

patrick-kidger:
LGTM - merged! :)

PR #114, opened 2024-11-17
Support C->R case

Randl:
Follow up on https://github.com/patrick-kidger/optimistix/issues/76

### DIFF
--- a/lineax/_misc.py
+++ b/lineax/_misc.py
@@ -113,6 +113,13 @@ def structure_equal(x, y) -> bool:
     return eqx.tree_equal(x, y) is True
 
 
+def is_complex_structure(structure):
+    return jnp.isdtype(
+        jnp.result_type(*(jax.tree.flatten(structure)[0])),
+        "complex floating",
+    )
+
+
 def complex_to_real_structure(in_structure):
     return jtu.tree_map(
         lambda x: ShapeDtypeStruct(
@@ -122,6 +129,17 @@ def complex_to_real_structure(in_structure):
         else x,
         in_structure,
     )
+
+
+def complex_to_real_tree(x, in_structure):
+    with jax.numpy_dtype_promotion("standard"):
+        return jtu.tree_map(
+            lambda x, struct: jnp.stack([x.real, x.imag], axis=-1)
+            if jnp.isdtype(struct.dtype, "complex floating")
+            else x,
+            x,
+            in_structure,
+        )
 
 
 def real_to_complex_tree(x, in_structure):
--- a/lineax/_misc.py
+++ b/lineax/_misc.py
@@ -19,6 +19,7 @@ import jax
 import jax.core
 import jax.numpy as jnp
 import jax.tree_util as jtu
+from jax import ShapeDtypeStruct
 from jaxtyping import Array, ArrayLike, Bool, PyTree  # pyright:ignore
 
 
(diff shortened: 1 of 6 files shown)


Randl:
@patrick-kidger I think this is a reasonable implementation of what we discussed. Some documentation changes are probably required. 
Mixed complex-real outputs are not supported here (but inputs are). We could aggressively assume that any real output means real output, but that would complicate implementation. It also would be a breaking change for cases where the function is properly linear (e.g., when complex and real inputs are multiplied each by a separate matrix). I don't think we can really detect this case and give it special treatment, but suggestions are welcome. All in all, keeping in mind the optimization use case, I think this is a reasonable degree of generality.
Also, if you see any edge cases I missed, I'll be happy to add them to the tests.

Randl:
ping @patrick-kidger

lineax/_misc.py
```diff
@@ -19,6 +19,7 @@
 import jax.core
 import jax.numpy as jnp
 import jax.tree_util as jtu
+from jax import ShapeDtypeStruct
```
patrick-kidger, line 22: Nit: I'd just `jax.ShapeDtypeStruct` directly for consistency with the rest of our imports.

lineax/_operator.py
```diff
@@ -1322,11 +1325,26 @@ def _(operator):
...
+        in_structure = operator.in_structure()
     flat, unravel = strip_weak_dtype(
-        eqx.filter_eval_shape(jfu.ravel_pytree, operator.in_structure())
+        eqx.filter_eval_shape(jfu.ravel_pytree, in_structure)
     )
+    fn = lambda x: operator.fn(map_to_original(unravel(x)))
     eye = jnp.eye(flat.size, dtype=flat.dtype)
-    jac = jax.vmap(lambda x: operator.fn(unravel(x)), out_axes=-1)(eye)
+
+    jac = jax.vmap(fn, out_axes=-1)(eye)
```
patrick-kidger, line 1347: After this change: given a C->R `FunctionLinearOperator`, call it `operator`, then I think `operator(some_complex_vector)` would have worked but `materialise(operator)(some_complex_vector)` would not work, as it now expects something from `R^2` instead?

The intention was that `materialise` would not change the observable input-output behaviour of an operator at all.

(FWIW I've just checked the current behaviour on `FunctionLinearOperator(lambda x: x.real, jax.ShapeDtypeStruct((), jnp.complex64))` and this is also wrong in the expected way: there's no way to express 'take a real part' when multiplying against a pytree, so it's not like the current state of affairs is any better... !)

It seems to me like the `AbstractLinearOperator` abstraction might actually just be fundamentally incompatible with complex dtypes, due to the not-really-defined nature of linear operators over such spaces?

patrick-kidger (commented):
Alright, sorry for the delay! I've left one main thought on this. WDYT?

lineax/_operator.py
Randl: You're right, as we discussed in https://github.com/patrick-kidger/optimistix/issues/76#issuecomment-2295215598 it is impossible to materialize `C->R` operator, since it is not complex linear but rather linear in imaginary and real parts. The two solutions I see is one in this PR (break the promise that materialize is noop) or just assert we do not support C->R operators (probably just give a warning) and suggest to user to make `R^2->R` operator out of it. The latter is definitely cleaner from the point of view of consistency, but may be less convenient to end user? Not sure about it, but luckily for me the decision is yours 🙃
The motivation was to support C->R stuff in optax so it can be a good testbed in terms of understanding how convenient this stuff is

patrick-kidger: I think we should follow JAX's lead on this one, with how `jax.jac{fwd,rev}` similarly error out for these kinds of operations.

So I think something like this might be the best choice, then!
```python
class AbstractLinearOperator(eqx.Module):
    def __check_init__(self):
        if is_complex_structure(self.in_structure()) and not is_complex_structure(self.out_structure()):
            raise ValueError(...)
```

In terms of end user convenience: I care a lot about this! My usual rule for this is that it is better not to support something than to support it awkwardly or with edge-cases. I think this leads to less frustration and a better UX overall. :)

Randl: Fair enough! Note that your solution will break existing cases for mixed-typed operators (i.e., single operator built from C->C and R->R blocks).

patrick-kidger: Hmm, that is a good point about block-with-mixed-type operators. Ach, complex support is very complicated! Is there a clean way to detect and allow that case, that you can see?

I am concerned that our complex support isn't quite meeting the above UX standards I'd like us to have... (this has definitely been a learning experience for me on how complex autodiff/operators/etc work!)

Randl: One possible way is to materialize the operator, but that can be expensive. I was thinking just issue a warning instead of exception.

Randl: Fair enough

Randl:
Replaced by https://github.com/patrick-kidger/optimistix/pull/235
Was outdated anyway

Randl:
Prohibiting C->R case at all could be still useful

PR #120, opened 2024-12-19
Adding a preconditioner example

vboussange:
Hey there, here is an attempt to make an example for the use of a preconditioner, as discussed in #8. The use of a Jacobi does not seem to provide much advantage though. @patrick-kidger Any idea of a low hanging fruit solution to have a nicer example? 

I personally would love to see an example with a hand crafted preconditioner such as the one proposed, and a more involved using e.g.`pyamg` preconditioning capability, with a possible `pure_callback` call.

### DIFF
--- a/docs/examples/preconditioner.ipynb
+++ b/docs/examples/preconditioner.ipynb
@@ -1,0 +1,195 @@
+{
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Using preconditioners\n",
+    "\n",
+    "\n",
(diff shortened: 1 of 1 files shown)


patrick-kidger:
In terms of the implementation, I like this example! It's short and easy to follow.

(Nit: `jr.PRNGKey -> jr.key`; the former has been replaced with the latter.)

Also tagging @frederic-bender from #8, as I don't have strong feelings about what would make for a stronger example of preconditioning. (I note that the [Julia folks](https://docs.sciml.ai/LinearSolve/stable/basics/Preconditioners/) also only consider a very simple example.)

Issue #124, opened 2025-01-02
Guidance for writing a `DiagonalPyTreeLinearOperator`


johannahaffner:
Hi Patrick, 

I'm implementing the interior-reflective Newton step method of Coleman and Li, for bound-constrained optimisation, in optimistix.
The core of this thing is to scale and translate the Hessian, such that

$B_k = D_k \cdot H_k \cdot D_k + diag(g_k) \cdot Jac(|v|) $

defines the operator for the Newton step, where the diagonal elements of $D_k$ are defined by a simple function on `y` (distance to bounds or $\pm1$ depending on some conditions on the gradient, all computed element-wise). For this to work for a general PyTree, I believe (?) the best thing to do would be to define a `DiagonalPyTreeLinearOperator`. 

I noticed that the Hessian approximation we're computing in `optx.BFGS` is actually "square", by which I mean that they do not have the redundant entries one finds, for example, in `jax.hessian`, and `as_matrix` returns a square matrix (this is not true for `jax.hessian`).  So maybe I'm making things more complicated than they need to be?

patrick-kidger:
So I think I can imagine two different kinds of 'diagonal pytree': one is where the pytree *is* the diagonal, and e.g. the diagonal array inside of `DiagonaLinearOperator` corresponds to its ravelled form. The second is where each element of the diagonal is a pytree, i.e. it's really a block diagonal operator.

If I understand your case correctly, I think you're in the first one? If so then I think that should be fairly straightforward to implement: pretty much copy-paste the existing `DiagonalLinearOperator` implementation, where the internal storage is now a pytree, `mv` is given by `tree_map(operator.mul, ...)` (I think), and `as_matrix` involves allocating a zero output matrix whose diagonal is filled in via `.at[].set()`. Does that make sense?

FWIW I'm not sure what you mean by 'redundant entries' in `jax.hessian`? Can you give an example?

johannahaffner:
> FWIW I'm not sure what you mean by 'redundant entries' in `jax.hessian`? Can you give an example?

Making an MWE I think I got to the bottom of this - I was taking a hessian of a function that accidentally returned a list of two values, rather than a scalar. This caused my matrix sizes to go to `(n, 2n)`, for `n` elements in a pytree. Since there were dictionaries in there, I was wondering if that caused a discrepancy from Schwarz' rule.

> If I understand your case correctly, I think you're in the first one? ... Does that make sense?

Yes and it does!

PR #125, opened 2025-01-04
Add a diagonal pytree linear operator

johannahaffner:
Adds a version of the diagonal linear operator that accepts and stores a PyTree.

Requesting main since dev does not exist at the moment :) 

As discussed in https://github.com/patrick-kidger/lineax/issues/124.

### DIFF
--- a/docs/api/operators.md
+++ b/docs/api/operators.md
@@ -28,6 +28,13 @@ Or, perhaps we only have a function $F : \mathbb{R}^m \to \mathbb{R}^n$ such tha
 
 ---
 
+::: lineax.DiagonalLinearOperator
+    selection: 
+        members: 
+            - __init__
+
+---
+
 ::: lineax.TridiagonalLinearOperator
     selection:
         members:
@@ -38,13 +45,6 @@ Or, perhaps we only have a function $F : \mathbb{R}^m \to \mathbb{R}^n$ such tha
 ::: lineax.PyTreeLinearOperator
     selection:
         members:
-            - __init__
-
----
-
-::: lineax.DiagonalLinearOperator
-    selection: 
-        members: 
             - __init__
 
 ---
--- a/docs/api/operators.md
+++ b/docs/api/operators.md
@@ -49,6 +49,13 @@ Or, perhaps we only have a function $F : \mathbb{R}^m \to \mathbb{R}^n$ such tha
 
 ---
 
+::: lineax.PyTreeDiagonalLinearOperator
+    selection: 
+        members: 
+            - __init__
+
+---
+
 ::: lineax.JacobianLinearOperator
     selection:
         members:
--- a/docs/api/operators.md
+++ b/docs/api/operators.md
@@ -28,13 +28,6 @@ Or, perhaps we only have a function $F : \mathbb{R}^m \to \mathbb{R}^n$ such tha
 
 ---
 
-::: lineax.DiagonalLinearOperator
-    selection:
-        members:
-            - __init__
-
----
-
 ::: lineax.TridiagonalLinearOperator
     selection:
         members:
@@ -49,7 +42,7 @@ Or, perhaps we only have a function $F : \mathbb{R}^m \to \mathbb{R}^n$ such tha
 
 ---
 
-::: lineax.PyTreeDiagonalLinearOperator
+::: lineax.DiagonalLinearOperator
     selection: 
         members: 
             - __init__
--- a/lineax/_operator.py
+++ b/lineax/_operator.py
@@ -1576,6 +1576,7 @@ def _(operator):
 
 @is_diagonal.register(IdentityLinearOperator)
 @is_diagonal.register(DiagonalLinearOperator)
+@is_diagonal.register(PyTreeDiagonalLinearOperator)
 def _(operator):
     return True
 
--- a/lineax/_operator.py
+++ b/lineax/_operator.py
@@ -485,38 +485,28 @@ class DiagonalLinearOperator(AbstractLinearOperator, strict=True):
     """
 
     diagonal: PyTree[Inexact[Array, "..."]]
-    input_structure: _FlatPyTree[jax.ShapeDtypeStruct] = eqx.field(static=True)
-    output_structure: _FlatPyTree[jax.ShapeDtypeStruct] = eqx.field(static=True)
 
     def __init__(self, diagonal: PyTree[ArrayLike]):
         """**Arguments:**
 
         - `diagonal`: an array or PyTree defining the diagonal of the matrix.
         """
-        diagonal = jtu.tree_map(inexact_asarray, diagonal)
-        structure = jax.eval_shape(lambda: diagonal)
-
-        self.diagonal = diagonal
-        self.input_structure = jtu.tree_flatten(structure)
-        self.output_structure = jtu.tree_flatten(structure)
+        self.diagonal = jtu.tree_map(inexact_asarray, diagonal)
 
     def mv(self, vector):
         return (ω(self.diagonal) * ω(vector)).ω
 
     def as_matrix(self):
-        array, _ = jfu.ravel_pytree(self.diagonal)
-        return jnp.diag(array)
+        return jnp.diag(diagonal(self))
 
     def transpose(self):
         return self
 
     def in_structure(self):
-        leaves, treedef = self.input_structure
-        return jtu.tree_unflatten(treedef, leaves)
+        return jax.eval_shape(lambda: self.diagonal)
 
     def out_structure(self):
-        leaves, treedef = self.output_structure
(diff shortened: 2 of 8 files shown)


patrick-kidger:
I think this pretty much LGTM! As it strictly generalizes `DiagonalLinearOperator` then WDYT about just updating the existing implementation (instead of creating a new operator)?

johannahaffner:
Oh good idea! This already works if the PyTree is just a vector.
I simply renamed it and removed the old DiagonalLinearOperator and all tests ran. 

I'll check later if there are other things, like supported addition operations (?) that need handling.

johannahaffner:
> I'll check later if there are other things, like supported addition operations (?) that need handling.

From reading the code, does not seem like there are any. To do a wider check for any unintentional breaks, I ran the test suites for diffrax and optimistix on my branch. No tests in either library are broken by this change.

lineax/_operator.py
```diff
@@ -474,6 +474,51 @@ def out_structure(self):
...
+
+    def __init__(self, pytree: PyTree[ArrayLike]):
+        """**Arguments:**
+
+        - `pytree`: a PyTree.
+        """
+        pytree = jtu.tree_map(inexact_asarray, pytree)
+        structure = jax.eval_shape(lambda: pytree)
+        
+        self.pytree = pytree
```
patrick-kidger, line 498: (Similarly I'd still call this `diagonal` -- we tend to name things by what they are, e.g. `solution.success`, not their type, e.g. `solution.boolean`.)

lineax/_operator.py
```diff
@@ -474,6 +474,51 @@ def out_structure(self):
...
+        self.input_structure = jtu.tree_flatten(structure)
+        self.output_structure = jtu.tree_flatten(structure)
+
+    def mv(self, vector):
+        return jtu.tree_map(lambda x, y: x * y, self.pytree, vector)
+
+    def as_matrix(self):
+        leaves = jtu.tree_leaves(self.pytree)
+        array = jnp.concatenate([jnp.asarray(leaf).ravel() for leaf in leaves])
```
patrick-kidger, line 507: `jax.flatten_util.ravel_pytree` might be better, as it also handles things like casting to a common dtype.

docs/api/operators.md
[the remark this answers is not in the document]
johannahaffner: I had changed the order to introduce it after `PyTreeLinearOperator`. Reversing now, since the description won't center pytrees.

lineax/_operator.py
```diff
@@ -474,6 +474,51 @@ def out_structure(self):
...
 
 
+class DiagonalLinearOperator(AbstractLinearOperator, strict=True):
+    """A diagonal linear operator, with a single PyTree defining the diagonal. It 
+    returns a PyTree of the same shape as the input PyTree. Only the PyTree is stored 
+    for efficiency, and only element-wise operations are performed.
+
+    This operator generalises a diagonal linear operator (the PyTree may be a 1D array), 
+    similar to how [lineax.PyTreeLinearOperator][] generalises a matrix linear operator.
+    """
```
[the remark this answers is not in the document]
johannahaffner: Like this?

lineax/_operator.py
johannahaffner: Done

lineax/_operator.py
```diff
@@ -474,6 +474,51 @@ def out_structure(self):
...
+
+    pytree: PyTree[Inexact[Array, "..."]]
+    input_structure: _FlatPyTree[jax.ShapeDtypeStruct] = eqx.field(static=True)
+    output_structure: _FlatPyTree[jax.ShapeDtypeStruct] = eqx.field(static=True)
+
+    def __init__(self, pytree: PyTree[ArrayLike]):
+        """**Arguments:**
+
+        - `pytree`: a PyTree.
```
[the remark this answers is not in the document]
johannahaffner: fixed

lineax/_operator.py
```diff
@@ -474,6 +474,51 @@ def out_structure(self):
...
+        """
+        pytree = jtu.tree_map(inexact_asarray, pytree)
+        structure = jax.eval_shape(lambda: pytree)
+        
+        self.pytree = pytree
+        self.input_structure = jtu.tree_flatten(structure)
+        self.output_structure = jtu.tree_flatten(structure)
+
+    def mv(self, vector):
+        return jtu.tree_map(lambda x, y: x * y, self.pytree, vector)
```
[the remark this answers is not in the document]
johannahaffner: Indeed much cleaner :)

lineax/_operator.py
```diff
@@ -1437,11 +1455,11 @@ def _(operator):
     return diagonal, off_diagonal, off_diagonal
 
 
-@tridiagonal.register(DiagonalLinearOperator)
-def _(operator):
-    (size,) = operator.diagonal.shape
-    off_diagonal = jnp.zeros(size - 1)
-    return operator.diagonal, off_diagonal, off_diagonal
+# @tridiagonal.register(DiagonalLinearOperator)
```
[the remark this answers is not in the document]
johannahaffner: Yup, missed that one!

lineax/_operator.py
```diff
@@ -1375,7 +1385,7 @@ def _(operator):
     return jnp.ones(operator.in_size())
 
 
-@diagonal.register(DiagonalLinearOperator)
+# @diagonal.register(DiagonalLinearOperator)
 @diagonal.register(TridiagonalLinearOperator)
 def _(operator):
     return operator.diagonal
```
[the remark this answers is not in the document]
johannahaffner: With a `filter_cond` based on `in_structure`?

lineax/_operator.py
johannahaffner: Didn't know about this one, this is much better.

lineax/_operator.py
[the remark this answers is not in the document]
johannahaffner: `cond` anything won't work because of the mismatched structure - so I wrote a simplistic if-else branch. Is that enough?

lineax/_operator.py
```diff
@@ -474,6 +474,51 @@ def out_structure(self):
...
+        """**Arguments:**
+
+        - `diagonal`: an array or PyTree defining the diagonal of the matrix.
+        """
+        diagonal = jtu.tree_map(inexact_asarray, diagonal)
+        structure = jax.eval_shape(lambda: diagonal)
+
+        self.diagonal = diagonal
+        self.input_structure = jtu.tree_flatten(structure)
+        self.output_structure = jtu.tree_flatten(structure)
```
patrick-kidger, line 501: I don't think we need to store these: we could just have
```python
def in_structure(self):
    return jax.eval_shape(lambda: self.diagonal)

def out_structure(self):
    return jax.eval_shape(lambda: self.diagonal)
```

lineax/_operator.py
```diff
@@ -1370,12 +1379,21 @@ def _(operator):
...
 
 
+@diagonal.register(DiagonalLinearOperator)
+def _(operator):
+    trivial_structure = jtu.tree_structure(jnp.ones(1,))  # diagonal is 1D array
+    is_trivial = jtu.tree_structure(operator.diagonal) == trivial_structure
+    if is_trivial:  # Avoid materialising the whole matrix in this case
+        return operator.diagonal
+    else:
+        return jnp.diag(operator.as_matrix())
```
patrick-kidger, line 1389: Actually, I take back my previous comment -- I think just this should work efficiently in all cases!
```python
diagonal, _ = jfu.ravel_pytree(operator.diagonal)
return diagonal
```

patrick-kidger: Honestly we could even implement `as_matrix()` in terms of it:
```python
def as_matrix(self):
    return jnp.diag(diagonal(self))
```

lineax/_operator.py
```diff
@@ -1429,6 +1447,14 @@ def _(operator):
     return diagonal, lower_diagonal, upper_diagonal
 
 
+@tridiagonal.register(DiagonalLinearOperator)
+def _(operator):
+    diagonal = jnp.diag(operator.as_matrix())
```
patrick-kidger, line 1452: This line could be `diagonal(operator)` to avoid materialising the whole matrix.

patrick-kidger (commented):
Okay, a few mores nits and I think we're there!

johannahaffner:
All fixed :)

lineax/_operator.py
johannahaffner: Much better!

patrick-kidger:
Wonderful stuff. LGTM, and merged :)

johannahaffner:
Thanks! You put a lot of thought into this, really appreciate it!

PR #129, opened 2025-01-13
Iterative Solvers return successfully when run with `max_steps` only

### LINKED ISSUE #126: Have iterative linear solvers run for a specified number of iterations
All iterative linear solvers contain this check in their constructor:

```python
        if isinstance(self.atol, (int, float)) and isinstance(self.rtol, (int, float)):
            if self.atol == 0 and self.rtol == 0 and self.max_steps is None:
                raise ValueError(
                    "Must specify `rtol`, `atol`, or `max_steps` (or some combination "
                    "of all three)."
                )
```

Hence, they can only be instantiated if at least one of `rtol`, `atol` or `max_steps` is prescribed. 

However, when **only** specifying `max_steps` (i.e., setting `rtol=0.0` and `atol=0.0`), the solution process will always throw an Equinox error:

```python
equinox.EquinoxRuntimeError: The maximum number of solver steps was reached. Try increasing `max_steps`.
```

This is because all iterative solvers will set the return flag to `RESULTS.max_steps_reached` if `steps==max_steps` (see, e.g., in CG [here](https://github.com/patrick-kidger/lineax/blob/936c31ccb4bfd4c2eec7b310b05431e61d197c67/lineax/_solver/cg.py#L218)). This is of course the desired behavior if one has set an additional tolerance threshold but I do have a usecase for which I only want a specific `max_steps` (and hence have the solver run for exactly that many steps). A quick hack is to set `throw=False` in `lineax.linear_solve(...)`. This deactivates the throw branch [here](https://github.com/patrick-kidger/lineax/blob/936c31ccb4bfd4c2eec7b310b05431e61d197c67/lineax/_solve.py#L103-L107), but also hinders a correct throw for other erroneous result types.

MWE:

```python
A = -2 * jnp.diag(jnp.ones(100)) + jnp.diag(jnp.ones(99), 1) + jnp.diag(jnp.ones(99), -1)
op = lineax.MatrixLinearOperator(A, tags=(lineax.negative_semidefinite_tag, lineax.symmetric_tag))
b = jax.random.normal(jax.random.key(0), (100,))

cg_solver = lineax.CG(0.0, 0.0, max_steps=5)
lineax.linear_solve(op, b, cg_solver)  # Raises an error
```

Ceyron:
A simple modification in the iterative linear solvers to not throw an `Equinox Error` if run with only a prescribed `max_steps`, e.g., `rtol=0.0, atol=0.0, max_steps=5`.

Closes #126.

### DIFF
--- a/lineax/_solver/bicgstab.py
+++ b/lineax/_solver/bicgstab.py
@@ -193,7 +193,7 @@ class BiCGStab(AbstractLinearSolver[_BiCGStabState], strict=True):
         else:
             result = RESULTS.where(
                 (num_steps == self.max_steps),
-                RESULTS.max_steps_reached,
+                RESULTS.max_steps_reached if has_scale else RESULTS.successful,
                 RESULTS.successful,
             )
         # breakdown is only an issue if we did not converge
--- a/lineax/_solver/cg.py
+++ b/lineax/_solver/cg.py
@@ -224,7 +224,7 @@ class _AbstractCG(AbstractLinearSolver[_CGState], strict=True):
         else:
             result = RESULTS.where(
                 num_steps == max_steps,
-                RESULTS.max_steps_reached,
+                RESULTS.max_steps_reached if has_scale else RESULTS.successful,
                 RESULTS.successful,
             )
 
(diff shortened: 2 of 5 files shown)


patrick-kidger:
LGTM -- and merged! Thank you for the improvement :)

Issue #81, opened 2024-02-08
On a failed solve, check whether the input was NaN as well


patrick-kidger:
Sometimes we detect that the output was `NaN`, and fail the linear solve. Under these circumstances we should check whether the input had any `NaN`s as well -- indicating that the culprit was probably something upstream.

Inspired by https://github.com/patrick-kidger/diffrax/issues/368.

PR #130, opened 2025-01-14
Check finite input

johannahaffner:
As described https://github.com/patrick-kidger/lineax/issues/81. (Came across this problem too.)

Will overwrite `RESULTS.singular` with a new `RESULTS.nonfinite_input` if the input is indeed nonfinite.


patrick-kidger:
LGTM!

Issue #116, opened 2024-11-20
Smoothing Methods like Jacobi & Gauss-Seidel


Ceyron:
Hi Patrick & Jason,

Thanks for this wonderful library. I was curious if you are interested in adding smoothing methods based on matrix decomposition, such as [the Jacobi method](https://en.wikipedia.org/wiki/Jacobi_method) or [Gauss-Seidel](https://en.wikipedia.org/wiki/Gauss%E2%80%93Seidel_method) (potentially with the option for relaxation).

Certainly, these solvers a way less efficient than the Krylov ones and need matrix materialization (the state of sparse matrices in JAX is not mature enough?!?), but they could complete the suite of solvers. They might also serve as a nice educational example.

vboussange:
Hey @Ceyron, this relates to https://github.com/patrick-kidger/lineax/issues/113 and https://github.com/patrick-kidger/lineax/issues/8 . Those methods would be very useful in multigrid methods that could be used for preconditioners to feed into e.g. GMRES. I am currently running a project that needs those. It could be nice to coordinate; I suggested to implement an example with preconditioning for solving a 2D Poisson system in https://github.com/patrick-kidger/lineax/issues/8, although I have not advanced on that yet.

Ceyron:
> I'd be happy to take a PR on these :)

Excellent. Then, we could start working on this in the coming weeks.

@vboussange: Do you already have some code for Jacobi or Gauss-Seidel? Otherwise, I would start looking into subclassing a new iterative solver since I need it for my research anyway.

vboussange:
@Ceyron Nope, I have not started digging into that - great that you can take the lead! I had in mind to get inspired from AMG.jl, specifically [here](https://github.com/JuliaLinearAlgebra/AlgebraicMultigrid.jl/blob/84e4a6a2fb0e3d43ad64b1945f0d199feac356c4/src/smoother.jl) and [there]( https://github.com/JuliaLinearAlgebra/AlgebraicMultigrid.jl/blob/84e4a6a2fb0e3d43ad64b1945f0d199feac356c4/src/multilevel.jl). Let me know if I can help!

Ceyron:
[...]

PR #131, opened 2025-01-15
Add Smoothers (Iterative Linear Solvers based on Matrix Decomposition)

Ceyron:
Hi @patrick-kidger,

This is to implement #116. I opened it as a draft because I wanted to get some feedback/have questions on handling Lineax internals and what features to add.

So far, I have implemented the `Jacobi` & `GaussSeidel` methods as a mixture of the dense solver (with their matrix materialization) and Krylov solvers (with their implementation of the iterative process). Moreover, I added a new tag `strictly_diagonally_dominant` because this is a sufficient (but not necessary) condition for their convergence (for [Jacobi](https://en.wikipedia.org/wiki/Jacobi_method#Convergence) and for [GaussSeidel](https://en.wikipedia.org/wiki/Gauss%E2%80%93Seidel_method#Convergence)). If `max_step=None`, the smoothers run for `100 * size` instead of `10 * size` (what the Krylov solvers default to) because of the slower convergence. I am thinking of bumping it to `200 * size` since I otherwise observe that sometimes some tests fail (due to the randomized matrices).

I have some questions to discuss before I polish up the PR with docs etc.:
1. Is my handling of the tag correctly? (with regards to `_operator.py` and `test_operator.py`) Do you think this new tag is a good addition? As far as I understand, Lineax does implement methods to *actually* check if operators obey their tags. It only relies on the user specification. (One could add a simple function to check if twice the absolute diagonal is larger than the absolute row sum.)
2. Are there additional tests for which I have to register the two new solvers?
3. I will implement the option for relaxation in both `Jacobi` and `GaussSeidel`. For the latter, this gives the [SOR method](https://en.wikipedia.org/wiki/Successive_over-relaxation). For symmetric matrices, there is also the [SSOR method](https://en.wikipedia.org/wiki/Symmetric_successive_over-relaxation). Do you think it is worth adding it as well?
4. Do you think that there should be some handling of the `strictly_diagonally_dominant` tag in the smoothing methods? Giving a warning, if it is not present? It is not a necessary condition, though.

These are the open points I will tackle after your feedback:

- [ ] Add Documentation
- [ ] Add a relaxation option to Jacobi and GaussSeidel and test it

### DIFF
--- a/lineax/__init__.py
+++ b/lineax/__init__.py
@@ -72,6 +72,7 @@ from ._tags import (
     lower_triangular_tag as lower_triangular_tag,
     negative_semidefinite_tag as negative_semidefinite_tag,
     positive_semidefinite_tag as positive_semidefinite_tag,
+    strictly_diagonally_dominant_tag as strictly_diagonally_dominant_tag,
     symmetric_tag as symmetric_tag,
     transpose_tags as transpose_tags,
     transpose_tags_rules as transpose_tags_rules,
--- a/lineax/__init__.py
+++ b/lineax/__init__.py
@@ -57,6 +57,7 @@ from ._solver import (
     CG as CG,
     Cholesky as Cholesky,
     Diagonal as Diagonal,
+    GaussSeidel as GaussSeidel,
     GMRES as GMRES,
     Jacobi as Jacobi,
     LU as LU,
--- a/lineax/__init__.py
+++ b/lineax/__init__.py
@@ -31,6 +31,7 @@ from ._operator import (
     is_lower_triangular as is_lower_triangular,
     is_negative_semidefinite as is_negative_semidefinite,
     is_positive_semidefinite as is_positive_semidefinite,
+    is_strictly_diagonally_dominant as is_strictly_diagonally_dominant,
     is_symmetric as is_symmetric,
     is_tridiagonal as is_tridiagonal,
     is_upper_triangular as is_upper_triangular,
--- a/lineax/__init__.py
+++ b/lineax/__init__.py
@@ -58,6 +58,7 @@ from ._solver import (
     Cholesky as Cholesky,
     Diagonal as Diagonal,
     GMRES as GMRES,
+    Jacobi as Jacobi,
     LU as LU,
     NormalCG as NormalCG,
     QR as QR,
--- a/lineax/_operator.py
+++ b/lineax/_operator.py
@@ -50,6 +50,7 @@ from ._tags import (
     lower_triangular_tag,
     negative_semidefinite_tag,
     positive_semidefinite_tag,
+    strictly_diagonally_dominant_tag,
     symmetric_tag,
     transpose_tags,
     tridiagonal_tag,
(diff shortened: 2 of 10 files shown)


patrick-kidger:
Thank you for putting this together! To work through your questions:

1. I think I'm inclined against this. As you note this is sufficient but not necessary -- e.g. [Sassenfeld's criterion](https://arxiv.org/abs/2201.05628) also suffices for the convergence of Gauss-Seidel -- and the tag system is pretty much entirely opt-in anyway.
2. I don't think so!
3. No strong feelings.
4. I think what would be best would be to arrange for a failed solve, together with a failed check of this sort, to produce a [RESULT](https://docs.kidger.site/lineax/api/solution/#lineax.RESULTS) that notes this and gives a more helpful error message.

Besides this, and as you note in #116, the use-cases for these are fairly unusual. C.f. also https://people.math.ethz.ch/~mhg/pub/biksm.pdf, which also highlights that Jacobi iterations are not usually a good way to do things. So I'm a little concerned about adding these just because of how not useful they are. I'm tagging @vboussange who also expressed interest in these in #116. Maybe these are okay to add as long as we add big warning signs in the documentation?

FWIW I am curious what your own interest is, both in this and in #132. My experience of iterative solvers is that they really don't work well for 'typical' dense matrices, whilst the larger sparse systems they're designed for are anyway fairly hard to make efficient in JAX.

vboussange:
Hey there, awesome work @Ceyron. I am myself quite interested by these solvers, as I would love to see a multigrid solver in JAX at some point. This work is an important milestone towards that.
However, matrix materialisation is a bummer for me. Could we avoid materialising matrices, by introducing custom `bcoo_tril` and `bcoo_triu`?

```python
@filter_jit
def bcoo_tril(mat: BCOO, k: int = 0) -> BCOO:
    """
    Return the upper-triangular part of the given 2D BCOO matrix.
    The result has zeros below the k-th diagonal.
    """
    rows = mat.indices[:, 0]
    cols = mat.indices[:, 1]
    mask = jnp.where(rows >= cols - k, 1.0, 0.0)
    new_data = mat.data * mask
    out = BCOO((new_data, mat.indices), shape=mat.shape)
    return out

@filter_jit
def bcoo_triu(mat: BCOO, k: int = 0) -> BCOO:
    """
    Return the upper-triangular part of the given 2D BCOO matrix.
    The result has zeros below the k-th diagonal.
    """
    rows = mat.indices[:, 0]
    cols = mat.indices[:, 1]
    mask = jnp.where(rows <= cols - k, 1.0, 0.0)
    new_data = mat.data * mask
    out = BCOO((new_data, mat.indices), shape=mat.shape)
    return out
    ```

Ceyron:
> Hey there, awesome work @Ceyron. I am myself quite interested by these solvers, as I would love to see a multigrid solver in JAX at some point.

Thanks for the kind words and the support :+1: 

>This work is an important milestone towards that. However, matrix materialisation is a bummer for me. Could we avoid materialising matrices, by introducing custom `bcoo_tril` and `bcoo_triu`?

Interesting, do you refer to using [`jax.experimental.sparse.BCOO`](https://docs.jax.dev/en/latest/_autosummary/jax.experimental.sparse.BCOO.html)? What is your experience with the experimental sparse features in JAX? Are they efficient? Is it a goal of the JAX team to eventually have them be part of stable JAX?

Ceyron:
Nice library :+1:.
I guess supporting BCOO sparse matrices in Lineax would require adding a new linear operator, like `lineax.BCOOLinearOperator`, which could potentially trigger more efficient solve paths in some solvers (like a sparse matrix decomposition for the smoothers or efficient sparse (incomplete) direct solvers (if JAX ever starts wrapping the corresponding [CuSparse routines](https://docs.nvidia.com/cuda/incomplete-lu-cholesky/index.html)). AFAIK, JAX does not support a way to translate a (linear) function into its sparse matrix representation like, e.g., [Julia has](https://github.com/JuliaDiff/SparseDiffTools.jl). Hence, if Lineax wants to support sparse matrices, they need to be directly supplied by the user. (Just leaving this pointer to [PhiFlow/PhiML](https://tum-pbs.github.io/PhiML/Matrices.html) that has a custom sparse matrix format and they find their sparsity pattern via tracing stencils). 

I suggest performing this pull request with dense matrix materialization (similar to what happens to any matrix-free linear operator in Lineax's direct solvers) and potentially adding sparse matrix support with a future update. I can imagine a sparse matrix support naturally leading to ways of implementing algebraic multigrid. Wdyt, @patrick-kidger?

patrick-kidger:
> Do you mean adding a new `RESULTS.non_diagonally_dominant`

Yup!

> I suggest performing this pull request with dense matrix materialization (similar to what happens to any matrix-free linear operator in Lineax's direct solvers) and potentially adding sparse matrix support with a future update.

Agreed. Hopefully (!) we can write the solvers in a matrix-free way that is agnostic to whether the operator is sparse or not. (Or even to what kind of sparse representation is being used -- BCOO, BCSR, ...)

PR #135, opened 2025-02-14
Include (semi-)definite in the description

johannahaffner:
Modifies the docstring to include the (semi-). Otherwise it sounds as though Cholesky is only picked for positive and negative definite operators.


patrick-kidger:
LGTM!

Issue #146, opened 2025-04-18
Can't run benchmarks


jpbrodrick89:
In drafting a PR to address #145 I noticed the benchmarks folder and thought it prudent to observe the results for `tridiagonal_solve`. Unfortunately, these no longer seem to run due to changes made to the code since these were introduced (namely `getkey` and `shaped_allclose` being removed/renamed in `tests/helpers`). I'm not sure whether it is desired for these to be revived or how they should be run (e.g. with `python`, `pytest` or `pytest-benchmark`).

johannahaffner:
Good morning!

And well spotted. I reckon that `shaped_allclose` used to be what `tree_allclose` was called. `getkey` is now defined in `tests/conftest.py`.

Hope that helps! If not, happy to take a look in the afternoon.

jpbrodrick89:
Thanks, yes `tree_allclose` works fine but as `getkey` is a fixture it can't be imported and called explicitly. I tried copying and pasting the definition into the file but Jax wasn't happy with the equinox key for some reason. I'm off for the Easter weekend so won't be able to share a copy of the error message until Tuesday unfortunately.

patrick-kidger:
Thanks for the report! I've just fixed these in https://github.com/patrick-kidger/lineax/commit/edc374f634bd62879e9975fab717209d7941f1a3 :)

PR #147, opened 2025-04-18
Improve Tridiagonal solve performance with primitive lowering

### LINKED ISSUE #145: Tridiagonal solve significantly slower than jax in-built when unbatched and for small batch sizes
The current implementation of tridiagonal solve in `lineax` uses a [manually implemented version](lineax/_solver/tridiagonal.py) of the Thomas algorithm with `jax.lax.scan` and a hard-coded `unroll` of 32. There is now a [`jax.linalg.tridiagonal_solve`](https://docs.jax.dev/en/latest/_autosummary/jax.lax.linalg.tridiagonal_solve.html) which directly calls the respective `cusparse` and `lapack` libraries when on GPU/CPU. JVP and transpose rules are supported on all devices [now](https://github.com/jax-ml/jax/pull/25787). However, despited the interface appearing batch friendly, timing scales linearly with batch size, whereas one would hope for sub-linear scaling on GPU. The latter is indeed achieved for the `lineax` implementation but the default unroll provides a sub-optimal starting point.

I have run tests for a simple tridiagonal solve of various arrays sizes (200–204,800) and it's VJP as well as various batch sizes for a fixed array size of 200. My main conclusions are
* Current (`unroll=32`) `lineax` implementation of forward solve is about 5x slower than `jax`'s respective implementation on same device.
* Current (`unroll=32`) `lineax` implementation becomes more efficient for forward solves when batch size is greater than ~20 (CPU) and ~5 (GPU).
* Current (`unroll=32`) `lineax` implementation of VJP is about 10x slower than `jax` on CPU and more than 200x slower on GPU.
* Fully rolled (`unroll=1`) `lineax` implementation of forward solve is comparable/slightly faster thsn `jax` `lapack` implementation on CPU.
* Fully rolled (`unroll=1`) `lineax` implementation of forward solve is faster than `jax` `lapack` implementation for batch sizes of 5 or more on CPU.
* Fully rolled (`unroll=1`) `lineax` implementation of VJP is still up to 5x slower for VJP.
* Fully **un**rolled (`unroll=200` initially) `lineax` is similar to `jax` on GPU (compilation time a lot longer of course).
* `unroll=200` is more efficient than `jax` for all batch sizes on GPU.
* Both rolled and fully unrolled implementations of VJP are about 100x slower on GPU.

I am quite confident my forward solve timings are fair and representative, but there is a chance my VJP timings might be off.

Options:
1. Expose setting of `unroll` to user with default value of 1 on CPU and 32 on GPU (maybe fully unrolled if possible but users might be surprised about long compilation time) and look into why VJP is running slower.
2. Just use `jax.linalg.tridiagonal_solve` and accept slow batch timings.
3. A combination of 1 and 2 where we use 1 for forward solves and 2 for VJP (and probably JVP too)

Happy to help make the PR once we decide on solution. Thanks!

Image

Image

Image

Tests run on a single Mac M2 CPU core and an NVIDIA A100 GPU on the NERSC Perlmutter supercomputer. To be specific, the problem solves the heat equation for a tanh temperature profile in a plasma with a thermal conductivity scaling as `T**2.5` and a relatively small timestep. The VJP is with respect to temperature with a Gaussian co-tangent vector.

Happy to share timing scripts but they required allowing customisation of unroll on my `lineax` fork and are also somewhat verbose as I wanted to ensure the matrix was well-posed and generalised well to different array sizes. Nevertheless, I confirm I 
- [x] Passed the diagonals as jax arrays across the jit boundary (operator created after)
- [x] Used `throw=False` for fair comparison
- [x] Used `block_until_ready` for representative timing. 
- [x] Called each function once with each set of input shapes before timing to eliminate compilation time.

jpbrodrick89:
WIP to fix #145 

Tridiagonal solve uses `jax.lax.linalg.tridiagonal_solve` when on GPU and the `lineax` implementation with `unroll=1` on CPU using JAX primitives. I think my `abstract_eval` is not general enough and am wondering whether I can get away with just importing `jax.lax.linalg.standard_linalg_primitive` (probably?)... If not, I'm undecided whether it's cleaner to have `abstract_eval` as a class method or a more verbosely named module-level function.

Done:
- [x] Lowering to CPU, GPU and other (unbatched)
- [x] Simple abstract eval
- [x] Single test for float64 (operator and vector) with array size of 200 unbatched on CPU and the expected speedup is observed.

TODO:
- [ ] Testing on GPU
- [ ] Implement a batching rule to use `lineax` implementation when batch size is greater than 1.
- [ ] Test batching
- [ ] Test mixed dtypes (if `lineax` intend to support that)
- [ ] Repeat my benchmarks
- [ ] Get `lineax` benchmarks running again (see #146)?

Sharing now as away for Easter weekend, will continue next week, feel free to share any interim feedback, thanks.

### DIFF
--- a/lineax/_solver/tridiagonal.py
+++ b/lineax/_solver/tridiagonal.py
@@ -12,15 +12,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from functools import partial
+import functools as ft
 from typing import Any
 from typing_extensions import TypeAlias
 
+import jax.extend as jex
+import jax.interpreters.mlir as mlir
 import jax.lax as lax
 import jax.numpy as jnp
-from jax._src.interpreters import mlir
 from jax.core import ShapedArray
-from jax.extend import core
 from jaxtyping import Array, PyTree
 
 from .._operator import AbstractLinearOperator, is_tridiagonal, tridiagonal
@@ -53,12 +53,6 @@ class Tridiagonal(AbstractLinearSolver[_TridiagonalState], strict=True):
                 "matrices"
             )
         return tridiagonal(operator), pack_structures(operator)
-
-    @classmethod
-    def abstract_eval(
-        cls, diagonal, lower_diagonal, upper_diagonal, vector, batch_unroll_gpu
-    ):
-        return ShapedArray(vector.shape, vector.dtype)
 
     def compute(
         self,
@@ -99,7 +93,13 @@ class Tridiagonal(AbstractLinearSolver[_TridiagonalState], strict=True):
         return False
 
 
-@partial(mlir.lower_fun, multiple_results=False)
+def _tridiagonal_solve_abstract_eval(
+    diagonal, lower_diagonal, upper_diagonal, vector, batch_unroll_gpu
+):
+    return ShapedArray(vector.shape, vector.dtype)
+
+
+@ft.partial(mlir.lower_fun, multiple_results=False)
 def _tridiagonal_solve_gpu_unbatched_lowering(
     diagonal, lower_diagonal, upper_diagonal, vector, _
 ):
@@ -111,7 +111,7 @@ def _tridiagonal_solve_gpu_unbatched_lowering(
     ).flatten()
 
 
-@partial(mlir.lower_fun, multiple_results=False)
+@ft.partial(mlir.lower_fun, multiple_results=False)
 def _tridiagonal_solve_cpu_lowering(
     diagonal, lower_diagonal, upper_diagonal, vector, _
 ):
@@ -158,9 +158,9 @@ def _tridiagonal_solve_lineax(diagonal, lower_diagonal, upper_diagonal, vector, 
     return solution
 
 
-_tridiagonal_solve_p = core.Primitive("_tridiagonal_solve_lineax")
+_tridiagonal_solve_p = jex.core.Primitive("_tridiagonal_solve_lineax")
 _tridiagonal_solve_p.def_impl(_tridiagonal_solve_lineax)
-_tridiagonal_solve_p.def_abstract_eval(Tridiagonal.abstract_eval)
+_tridiagonal_solve_p.def_abstract_eval(_tridiagonal_solve_abstract_eval)
 
 mlir.register_lowering(
     _tridiagonal_solve_p, _tridiagonal_solve_cpu_lowering, platform="cpu"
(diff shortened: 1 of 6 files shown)


lineax/_solver/tridiagonal.py
```diff
@@ -12,11 +12,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from functools import partial
```
patrick-kidger, line 15: Style nit, `import functools as ft; ft.partial` is the style I used throughout the Equinox ecosystem.

lineax/_solver/tridiagonal.py
```diff
@@ -12,11 +12,15 @@
...
 from typing import Any
 from typing_extensions import TypeAlias
 
 import jax.lax as lax
 import jax.numpy as jnp
+from jax._src.interpreters import mlir
```
patrick-kidger, line 21: Can we do `import jax.interpreters.mlir as mlir` instead? (This is both a style comment, and a question on whether the public API is enough?)

lineax/_solver/tridiagonal.py
```diff
@@ -12,11 +12,15 @@
...
+from jax.core import ShapedArray
+from jax.extend import core
```
patrick-kidger, line 23: Likewise style comment, `import jax.extend as jex; jex.core.Foo`.

lineax/_solver/tridiagonal.py
```diff
@@ -50,51 +54,29 @@ def init(self, operator: AbstractLinearOperator, options: dict[str, Any]):
             )
         return tridiagonal(operator), pack_structures(operator)
 
+    @classmethod
+    def abstract_eval(
+        cls, diagonal, lower_diagonal, upper_diagonal, vector, batch_unroll_gpu
+    ):
+        return ShapedArray(vector.shape, vector.dtype)
```
patrick-kidger, line 61: I agree with your comment, this should be a global function!

As for its generality, I think it's basically fine -- we perform the solve on the ravelled form of the input, so just a single array is what's expected.

lineax/_solver/tridiagonal.py
```diff
@@ -117,6 +99,83 @@ def allow_dependent_rows(self, operator):
...
+)
+mlir.register_lowering(
+    _tridiagonal_solve_p, _tridiagonal_solve_gpu_unbatched_lowering, platform="cuda"
+)
+mlir.register_lowering(
+    _tridiagonal_solve_p, _tridiagonal_solve_gpu_unbatched_lowering, platform="rocm"
+)
+mlir.register_lowering(
+    _tridiagonal_solve_p,
+    mlir.lower_fun(_tridiagonal_solve_lineax, multiple_results=False),
```
patrick-kidger, line 176: Style nit, `mlir.lower_fun` is used as a decorator on the other branches and as an inline call here.

patrick-kidger (commented):
Thank you for putting this draft together! I've left some quick comments.

> whether I can get away with just importing `jax.lax.linalg.standard_linalg_primitive`

Best we don't I think, that sounds like exactly the kind of obscure thing they'll deprecate a year from now! :grin:

PR #151, opened 2025-05-07
Added jax tridiagonal solver benchmark

jpbrodrick89:
Currently only the `lineax` Tridiagonal solver was used without a comparison to `jax`. This is addressed in this PR. However, due to 50 being a very small matrix size for meaningful tridiagonal benchmarks the results are quite noisy. The jax implementation bottoms out at around 1E-5 for an array size of 200. When benchmarks are more refined we should probably use different size benchmarks for different solvers (Cholesky is very slow and struggles immensely with a size of 200).


patrick-kidger:
Awesome! This makes sense to me as an addition. Merged :)

PR #152, opened 2025-05-14
Improve tridiagonal solve performance without additional primitives

jpbrodrick89:
Now that [batched tridiagonal solves use cusparse batched](https://github.com/jax-ml/jax/issues/28544) in `jax.lax.tridiagonal_solve` (nightly release only for now!!) we don't really need to use fancy lowering for our tridiagonal solve. In this PR `Tridiagonal` _always_ uses `jax.lax.tridiagonal_solve`. If the faster but unstable Thomas algorithm is desired this is now provided by the `Thomas` solver, which has `unroll=1` for optimal CPU performance (at the cost of terrible GPU performance). In the future we can look to improve the Thomas algorithm runtime on GPU by lowering to `cuThomas`.

This might be ready to go, as benchmarks run fine (haven't run tests yet as they take a while) but let me know if I've missed someone or we want to change interface or make it easier to use `Thomas`.

### DIFF
--- a/lineax/_solver/tridiagonal.py
+++ b/lineax/_solver/tridiagonal.py
@@ -35,7 +35,11 @@ _TridiagonalState: TypeAlias = tuple[tuple[Array, Array, Array], PackedStructure
 
 
 class Tridiagonal(AbstractLinearSolver[_TridiagonalState], strict=True):
-    """Tridiagonal solver for linear systems, using the Thomas algorithm."""
+    """Tridiagonal solver for linear systems, using the LAPACK/cusparse implementation
+    of Gaussian elimination with partial pivotting. This is typically slightly slower
+    on CPU than the Thomas algorithm, but is unconditionally stable and still exhibits
+    linear scaling with the size of the system.
+    ."""
 
     def init(self, operator: AbstractLinearOperator, options: dict[str, Any]):
         del options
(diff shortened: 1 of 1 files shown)


patrick-kidger:
Nice! That's super simple.

Do you think it's even worth keeping the `Thomas` solver around? We could just remove it.

lineax/_solver/tridiagonal.py
```diff
@@ -50,6 +54,58 @@ def init(self, operator: AbstractLinearOperator, options: dict[str, Any]):
...
+        return False
+
+    def allow_dependent_rows(self, operator):
+        return False
+
+
+class Thomas(Tridiagonal):
+    """Thomas algorithm for tridiagonal linear systems. This is typically faster on
+    CPU than the more stable LAPACK/cusparse implementation of Gaussian elimination
+    with partial pivotting used in `Tridiagonal`. It will currently be much slower
```
johannahaffner, line 100: Nit: I think pivoting has just one t. Irrelevant if `Thomas` gets removed anyway.

johannahaffner (commented):
Typo, potentially irrelevant :) also: nice work!

lineax/_solver/tridiagonal.py
jpbrodrick89: Good catch! Will fix if we keep.

jpbrodrick89:
More than happy to remove as agree with the "footgun" argument 🙂 , but maybe the right compromise for now is to just remove it from public API (base `__init__.py` and the docs) and keep the source code (`_solvers/tridiagonal.py` and `_solver/__init__.py`) in case we change our mind to support experienced power users? I don't think I will use it for my use case for now (stabiiity seems to be an issue but not sure whether it's tridiagonal solves or something else!) but it seems the Thomas algorithm on CPU can be up to 20% faster unbatched and 2x faster for some batch sizes which some users could find value.

Would you want me to edit the docstrings at all in either case?

jpbrodrick89:
For merging into main, is your plan to merge once we're converged and addressed the above but delay cutting a new release until the next Jax PyPi release?

lineax/_solver/tridiagonal.py
```diff
@@ -50,6 +54,58 @@ def init(self, operator: AbstractLinearOperator, options: dict[str, Any]):
...
+        return conj_state, options
+
+    def allow_dependent_columns(self, operator):
+        return False
+
+    def allow_dependent_rows(self, operator):
+        return False
+
+
+class Thomas(Tridiagonal):
```
patrick-kidger, line 97: Oh, I've just clocked that this is inheriting from the concrete class. We don't do that around here! [link](https://docs.kidger.site/equinox/pattern/)

I think just inheriting from `AbstractLinearSolver` is all that is needed here.

patrick-kidger (commented):
> remove it from public API (base __init__.py and the docs) and keep the source code and private API

That sounds reasonable to me!

> Would you want me to edit the docstrings at all in either case?

I think the current docstrings seem fine, IIUC!

> For merging into main, is your plan to merge once we're converged and addressed the above but delay cutting a new release until the next Jax PyPi release?

Yup I think that sounds reasonable.

lineax/_solver/tridiagonal.py
jpbrodrick89: Thanks, that makes a lot of sense, thanks for the link, I had the feeling I was creating a code smell! I think I will need to introduce an `AbstractTridiagonalSolver` to avoid duplication of methods other than `compute` right?

patrick-kidger: If so then I think that probably speaks in favour of just deleting it! It can continue to exist on this branch if we need to resurrect it later.

jpbrodrick89:
Removed the Thomas algoirthm and simplified docstring accordingly, very simple diff now! :)

jpbrodrick89:
That's not a problem as we can always checkout the fork instead, thanks Patrick! 😃  As a reminder, this PR will still 'work' with the current JAX version it will just be slower for large batch sizes until the new JAX version

PR #158, opened 2025-06-09
Replace allow_dependent_rows and allow_dependent_columns with assume_full_rank

adconner:
The two functions allow_dependent_{rows,columns} together did the job of answering if the solver accepts full rank matrices for the purposes of the jvp. Allowing them to be implemented separately created some issues: 
1) Invalid states were representable. Eg. What does it mean that
   dependent columns are allowed for square matrices if dependent rows
   are not? What does it mean that dependent rows are not allowed for
   matrices with more rows than columns?
2) As the functions accept operator as input, a custom solver could in
   principle decide its answer based on operator's dynamic value rather
   than only jax compilation static information regarding it, as in all
   the lineax defined solvers. This would prevent jax compilation and
   jit.

Both issues are addressed by asking the solver to report only if it assumes the input is numerically full rank. If this assumption is exactly violated, its behavior is allowed to be undefined, and is allowed to error, produce NaN values, and produce invalid values.

### DIFF
--- a/lineax/_solver/qr.py
+++ b/lineax/_solver/qr.py
@@ -100,22 +100,8 @@ class QR(AbstractLinearSolver):
         conj_options = {}
         return conj_state, conj_options
 
-    def allow_dependent_columns(self, operator):
-        rows = operator.out_size()
-        columns = operator.in_size()
-        # We're able to pull an efficiency trick here.
-        #
-        # As we don't use a rank-revealing implementation, then we always require that
-        # the operator have full rank.
-        #
-        # So if we have columns <= rows, then we know that all our columns are linearly
-        # independent. We can return `False` and get a computationally cheaper jvp rule.
-        return columns > rows
-
-    def allow_dependent_rows(self, operator):
-        rows = operator.out_size()
-        columns = operator.in_size()
-        return rows > columns
+    def assume_full_rank(self):
+        return True
 
 
 QR.__init__.__doc__ = """**Arguments:**
--- a/docs/api/solvers.md
+++ b/docs/api/solvers.md
@@ -9,9 +9,9 @@ If you're not sure what to use, then pick [`lineax.AutoLinearSolver`][] and it w
             members:
                 - init
                 - compute
-                - allow_dependent_columns
-                - allow_dependent_rows
                 - transpose
+                - conj
+                - assume_full_rank
 
 ::: lineax.AutoLinearSolver
     options:
--- a/lineax/_solve.py
+++ b/lineax/_solve.py
@@ -198,16 +198,17 @@ def _linear_solve_jvp(primals, tangents):
         # -A'x term
         vec = (-(t_operator.mv(solution) ** ω)).ω
         vecs.append(vec)
-        allow_dependent_rows = solver.allow_dependent_rows(operator)
-        allow_dependent_columns = solver.allow_dependent_columns(operator)
-        if allow_dependent_rows or allow_dependent_columns:
+        rows, columns = operator.out_size(), operator.in_size()
+        assume_independent_rows = solver.assume_full_rank() and rows <= columns
+        assume_independent_columns = solver.assume_full_rank() and columns <= rows
+        if not assume_independent_rows or not assume_independent_columns:
             operator_conj_transpose = conj(operator).transpose()
             t_operator_conj_transpose = conj(t_operator).transpose()
             state_conj, options_conj = solver.conj(state, options)
             state_conj_transpose, options_conj_transpose = solver.transpose(
                 state_conj, options_conj
             )
-        if allow_dependent_rows:
+        if not assume_independent_rows:
             lst_sqr_diff = (vector**ω - operator.mv(solution) ** ω).ω
             tmp = t_operator_conj_transpose.mv(lst_sqr_diff)  # pyright: ignore
             tmp, _, _ = eqxi.filter_primitive_bind(
@@ -221,7 +222,7 @@ def _linear_solve_jvp(primals, tangents):
             )
             vecs.append(tmp)
 
-        if allow_dependent_columns:
+        if not assume_independent_columns:
             tmp1, _, _ = eqxi.filter_primitive_bind(
                 linear_solve_p,
                 operator_conj_transpose,  # pyright: ignore
@@ -393,56 +394,6 @@ class AbstractLinearSolver(eqx.Module, Generic[_SolverState]):
         """
 
     @abc.abstractmethod
-    def allow_dependent_columns(self, operator: AbstractLinearOperator) -> bool:
-        """Does this method ever produce non-NaN outputs for operators with linearly
-        dependent columns? (Even if only sometimes.)
-
-        If `True` then a more expensive backward pass is needed, to account for the
-        extra generality.
-
-        If you do not need to autodifferentiate through a custom linear solver then you
-        simply define this method as
-        ```python
-        class MyLinearSolver(AbstractLinearsolver):
-            def allow_dependent_columns(self, operator):
-                raise NotImplementedError
-        ```
-
-        **Arguments:**
-
-        - `operator`: a linear operator.
-
-        **Returns:**
-
-        Either `True` or `False`.
-        """
-
-    @abc.abstractmethod
-    def allow_dependent_rows(self, operator: AbstractLinearOperator) -> bool:
-        """Does this method ever produce non-NaN outputs for operators with
-        linearly dependent rows? (Even if only sometimes)
-
-        If `True` then a more expensive backward pass is needed, to account for the
-        extra generality.
-
-        If you do not need to autodifferentiate through a custom linear solver then you
-        simply define this method as
-        ```python
-        class MyLinearSolver(AbstractLinearsolver):
-            def allow_dependent_rows(self, operator):
-                raise NotImplementedError
-        ```
-
-        **Arguments:**
-
-        - `operator`: a linear operator.
-
-        **Returns:**
-
-        Either `True` or `False`.
-        """
-
-    @abc.abstractmethod
     def transpose(
         self, state: _SolverState, options: dict[str, Any]
     ) -> tuple[_SolverState, dict[str, Any]]:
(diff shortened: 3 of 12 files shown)


adconner:
If you wanted to go further than this PR and did not mind a breaking change, you can make full_rank an operator tag. Then solvers requiring full_rank operators check this just like they check any tag, and the jvp checks the operator tag when choosing the jvp terms to emit. This has benefits

1. More general solvers also benefit from a faster jvp when called with operators statically known (or assumed) to be full rank
2. AutoLinearSolver can now be fully automatic, not requiring well_posed for an argument
3. It makes all static assumptions about operators needed by solvers explicit in operator tags

Users would have to update their code to tag most operators as full rank

patrick-kidger:
Hey there! So you're actually touching on a design point from way back in the early days of Lineax. We in fact originally had a `maybe_singular_tag` on the operator rather than a description of what could be handled by the solver.

We eventually nixed this because it was too much of a footgun.
- If Lineax defaulted to assuming well-posedness and you left off the tag `maybe_singular_tag` then you'd silently incorrect JVPs.
- If Lineax defaulted to assuming singularity (as you seem to be suggesting here) and you left off the `full_rank` tag then you'd get silently expensive JVPs. Or conversely if adding that tag because the accepted default, then forgetting to remove it puts you back in the land of silently incorrect JVPs.

Either way this was clearly going to go wrong for a lot of users a substantial fraction of the time!

Making this a property of the solver instead reflects real-world usage. You're probably only using SVD if you have a singular problem, for example, and you're anyway happy to accept the extra computational cost in the JVP rule.

---

As for the adjustment you're originally making in this PR: being able to distinguish these two cases is useful for the sake of faster JVPs, in particular for QR solves.

On (1) it's true that illegal states are representable. We could probably add some `assert` statements to `_linear_solve_jvp` to catch the cases you highlight, which are all determinable given compiletime information. On (2) indeed it has to be static; this is already captured in its type annotation `-> bool` which is not a JAX array; it is an error for a solver to return anything else -- as meaningless as returning a string or `object()` etc.

WDYT?

patrick-kidger:
> It is true that AutoLinearSolver would be SVD for undecorated operators, but this is the only source of silent inefficiencies (and this is also analagous to the psd tag: the user will already get silent inefficiencies from the auto solver selection if they do not tag a known psd operator)

Actually, there is one further problem here (and one that is a motivating reason for us not to do this ): if a user has what they believe to be a definite operate, but mistakenly provides an indefinite operator, then SVD will silently compute a pseudoinverse (least-square/least-norm) solve, rather than erroring out as would be desirable.

> Any values of allow_dependent_{rows,columns} which are not functions of the operator being {tall,square,wide} and whether it is assumed to be full rank are invalid.

On this part, I think I like these changes. This was always a subtle point I have to think a lot about.

I think my main comment is that in general I think this could still be a function of any structure/sparsity in the operator? I'd need to noodle on this to be sure I'm getting it right, but I could believe that certain solvers might exhibit or not exhibit this behaviour depending on the operator. (The trivial example is of course a user-defined `assume_full_rank` tag that the solver checks and respects! But perhaps there are other examples?)

patrick-kidger:
> so the the only information it is missing is whether the operator is full rank

So let's consider a solver that consumes a `2n`x`n` matrix and a `2n`-length vector and returns an `n`-length solution, and for which `assume_full_rank = False`; that is to say it will always return a pseudoinverse solution.

I've not yet told you anything about the structure of the matrix or the behaviour of the solver.

According to the description you've given here, we would always compute a more-expensive JVP, corresponding to the allow-dependent-rows case.

However! It just so happens that I have a case in mind for which we may only need the inexpensive JVP. (The part corresponding to a true linsolve solution, not the more general pseudoinverse solution).

Namely, we introduce all of the following:
```python
# Consumes two n x n operators and stacks them to create a 2n x n operator.
class TwoOperatorsInATrenchcoast(AbstractLinearOperator):
    first: AbstractLinearOperator
    second: AbstractLinearOperator

    ...

# Matrix of all zeros
class ZeroLinearOperator(AbstractLinearOperator):
    ...

def is_full_and_zero(operator: AbstractLinearOperator) -> bool:
    return (isinstance(operator, TwoOperatorsInATrenchcoast) and isinstance(operator.second, ZeroLinearOperator))

class WeirdSolver(AbstractLinearSolver):
    ...

    def allow_dependent_rows(self, operator):
        return not is_full_and_zero(operator)

    def compute(self, ...):
        if is_full_and_zero(operator) and is_lower_half_zero(vector):  # second function elided for brevity, can be statically evaluatable if there is suitable pytree structure (just as with the operator).
            # Perform an LU solve on the upper half, then pad with zeros
        else:
            # Perform an SVD solve
```

In this case we have that `WeirdSolver` is indeed satisfying its contract of only returning a pseudoinverse solution. In the fastpath branch, this coincides with performing a full-rank solve on a submatrix. When on that branch we can skip the extra JVP computation (the tangents in the zero regions are also structurally zero), and this is reflected in `allow_dependent_rows`.

Thus we have obtained a case in which the solver needs to consume the operator to determine the answer to how it handles dependent rows orn ot.

---

The above is obviously highly contrived. But I think it indicates that completely factoring out the operator height-vs-width directly into the JVP rule would indeed lose expressivity.

(Now, one could certainly make the argument that this is a worthy trade-off in the name of simplicity. And if we done this the first time around I'd agree, but as it is I'm inclined to keep things for the sake of backwards-compatibility.)

WDYT?

patrick-kidger:
Sorry for taking so long to get back to you; as this is a very technical + mostly internal change, it's been pushed down my priority list.

Anyway, I think we agree with each other then! As per my last message:

> (Now, one could certainly make the argument that this is a worthy trade-off in the name of simplicity. And if we done this the first time around I'd agree, but as it is I'm inclined to keep things for the sake of backwards-compatibility.)

Indeed a simpler solution could have been arrived at, at the cost of an acceptable loss of generality. However breaking backward compatibility requires a very high bar, and I don't think that adjusting where we land on this trade-off meets that.

I am concious of your other work in #159. I'd definitely still be very happy to get that in if you'd be willing to update it without this PR as a dependency?

lineax/_solver/qr.py
```diff
@@ -100,22 +100,8 @@ def conj(self, state: _QRState, options: dict[str, Any]):
...
-        # So if we have columns <= rows, then we know that all our columns are linearly
-        # independent. We can return `False` and get a computationally cheaper jvp rule.
-        return columns > rows
-
-    def allow_dependent_rows(self, operator):
-        rows = operator.out_size()
-        columns = operator.in_size()
-        return rows > columns
+    def assume_full_rank(self):
+        return True
```
patrick-kidger, line 104: I think this one is a bit of a footgun. When using `AutoLinearSolver(well_posed=None)` then we'll select QR if the operator is nonsquare, but as your example in #168 demonstrates, then this may erroneously return a solution.

Unlike when we first wrote Lineax, I believe JAX now has a rank-revealing QR decomposition available via [`jax.lax.linalg.qr`](https://docs.jax.dev/en/latest/_autosummary/jax.lax.linalg.qr.html).

I think setting the return value here to `False`, and unconditionally using the rank-revealing implementation (modulo a 'promise full rank' tag, more on that later), should fix this however?

patrick-kidger (commented):
Okay, on the basis of #168 - returning to this PR @adconner, I think I'm now open to merging this more-or-less as-is. The convincing factor for me was the difficulty of getting `allow_dependent_{rows,columns}` to be correct, as you highlighted a long-standing bug in the QR solver of exactly this sort.

I have one question on QR below, otherwise I like this PR as-is.

lineax/_solver/qr.py
adconner: All the solvers except SVD which are assume_full_rank=True might erroneously return a solution in the rank deficient case. If you wish to test this use a 3x3 matrix with the bottom right 2x2 zero and try LU, or make it symmetric and try Cholesky. This is just a fact of life, if you need guaranteed errors you use SVD or pivoted qr.

In my mind, based on the solver selection in the documentation, `AutoLinearSolver(well_posed=None)` is supposed to mean "handle full rank non square operators", `True` means "handle arbitrary non square operators" and `False` means "handle full rank square operators". The language in the documentation is a bit indirect, but seems to support this ("ill posed" meaning "nonsquare", fine, like "over/underdetermined"). Maybe this isn't what you had in mind when you implemented this option originally but this is the actual interface lineax currently provides.

And as an aside, you *definitely* want to efficiently explicitly handle full rank nonsquare operators, both with solver and grad support, as this is almost the most common case: Remember than a random matrix is full rank, so also are most matrices encountered "in the wild". The ways you might encounter rank deficiency are (1) if your application really has it for some structural reason, ie all matrices in sight have the same rank which is not full, in which case you probably know or you find out during development by checking the singular values of a random example from the application, or (2) you are optimizing in the space of full rank matrices and your objective takes you close to rank deficient matrices. In case (1) you want SVD or pivoted qr, and in case (2) you probably *dont* want to start suddenly changing the structure of the objective function, computing minimum norm solutions as your singular values get small (not to mention changing your gradients, which are only valid in a locus of constant rank matrices, recall e.g. the first sentence of https://en.wikipedia.org/wiki/Moore-Penrose_inverse#Derivative ). In this case you probably don't want it hidden from you that you are close to dividing by zero. (As a truly tangential aside, in case (1) you have also the same concern as case (2): maybe you are optimizing in the space of constant rank r matrices and you get close to the rank <r-1 locus. You probably dont want your solver to suddenly start assuming an extra singular value is zero. Currently there is no clear way to work around this aside from forking lineax to add operator rank assumptions.)

Its also for exactly this case (full rank non square) that I contributed Normal, so that we can have many more solvers in this space, most simply Normal(Cholesky). 

Aside from the problem of wanting to ignore the above use case, there is another subtlety implementing rank revealing qr only and making it `assume_full_rank=False`. Firstly, its relatively simple to implement it to solve full rank systems with guaranteed errors on rank deficiency if desired (at least it would only have the same problems as the current qr, see jax-ml/jax#29173 ). This would still be `assume_full_rank=True`, but with guaranteed checking of this assumption at run time. However, a version doing a pseudo inverse solve needs more lapack (and equivalent cuda/tpu) routines than currently provided by jax, to do the so-called "complete orthogonal factorization": see the documentation of [sgelsy](https://www.netlib.org/lapack/explore-html/dc/d8b/group__gelsy_gada8d9d36935b71771a6bf91ca57da6ad.html). The additional needed routine ([tzrzf](https://www.netlib.org/lapack/explore-html/dd/d82/group__tzrzf_ga1137e671067bad30ac49c0735caaefdd.html)) is not provided by jax; I also mention this in the last paragraph of that jax issue.

In any case, even if you want to add the required support to jax to implement full pseudo inverse pivoted qr in lineax, you would still want to support pivoted and ordinary qr for assumed full rank matrices, because there is a real saving on avoiding a jvp term and this is the common case (and in the pivoted qr case, you additionally save the step of forming the complete orthogonal decomposition).

PR #159, opened 2025-06-09
Implement Normal, a solver applying an inner solver to the normal equations

### LINKED ISSUE #157: Bug: NormalCG jvp can do CG on singular systems
NormalCG applies CG to the normal equations A* A x = A* b, and CG's documentation requires A* A to be nonsingular. This is the case if A is full rank and tall, but is not if A is wide, even if A is full rank. The problem is the jvp requires a normalcg solve with A replaced by A*, which is wide if A is tall.

The fix is simple: the issue is only that the wrong normal equations are used for the wide case, here they are A A^* y = b, where x = A^* y. In the wide A case, x is the minimum norm solution, and in the tall case, x is the solution in the least squares sense. In both cases, x is the pseudoinverse solution. Fix should be trivial

See also #156

adconner:
This fixes #157 by implementing what we discussed there, a wrapper which takes any solver and applies it to the normal equations.

I've stacked this change atop #158 as the implementation is significantly cleaner taken with the change there.

This currently includes the breaking change of removing `lx.NormalCG(...)` in favor of `lx.Normal(lx.CG(...))`, so perhaps a simple definition should be made for backward compatibility. It was not clear how I should handle that from the standpoint of the documentation, as the following suffices

```python
def NormalCG(rtol: float,
        atol: float,
        norm: Callable[[PyTree], Scalar] = max_norm,
        stabilise_every: int | None = 10,
        max_steps: int | None = None):
    return Normal(CG(rtol,atol,norm,stabilise_every,max_steps))
```

but now the name NormalCG is a function, not a class, and its not clear if it be documented along with the other solvers. Would you want to add the above as an alternative, possibly deprecated interface? And if so should it be documented, and if so how?

### DIFF
--- a/docs/api/solvers.md
+++ b/docs/api/solvers.md
@@ -9,9 +9,9 @@ If you're not sure what to use, then pick [`lineax.AutoLinearSolver`][] and it w
             members:
                 - init
                 - compute
-                - allow_dependent_columns
-                - allow_dependent_rows
                 - transpose
+                - conj
+                - assume_full_rank
 
 ::: lineax.AutoLinearSolver
     options:
--- a/docs/api/solvers.md
+++ b/docs/api/solvers.md
@@ -42,9 +42,16 @@ These are capable of solving ill-posed linear problems.
         members:
             - __init__
 
+---
+
+::: lineax.Normal
+    options:
+        members:
+            - __init__
+
 !!! info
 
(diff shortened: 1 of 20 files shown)


patrick-kidger:
Looks like the test failures are from #172 and unrelated, so merging!

Issue #160, opened 2025-06-13
Preconditioner for transposed krylov methods


f0uriest:
I'm using `lx.GMRES` with a preconditioner and trying to understand why the forward solve converges fine but the gradient doesn't. Looking through the source, it seems like the `solver.transpose` throws away the preconditioner that's passed in `options`:

https://github.com/patrick-kidger/lineax/blob/66b7b5327a44e4b944a8ce9242773150e8a8d811/lineax/_solver/gmres.py#L416-L420

So the tangent solve here:
https://github.com/patrick-kidger/lineax/blob/66b7b5327a44e4b944a8ce9242773150e8a8d811/lineax/_solve.py#L285-L294

Is done without preconditioning, which seems wrong. Am I misreading something?

patrick-kidger:
Hmm, throwing away the preconditioner is probably a mistake. Happy to take a PR on that! (And ideally a test 😄)

As for your second comment, I'm away from my laptop right now but I agree this seems not desirable. I think `GMRES` has been one of our less-loved children so it's possible that we've missed something in its implementation.

f0uriest:
Sure I'll try to put together a PR for the preconditioner issue this week, and also try to fix some of the issues from #139 

For the termination criteria, here's a more complete example:
```python
# random nonsymmetric poorly conditioned matrix
A = np.random.random((10,10))
A += np.diag(np.arange(10)**6)

b = np.random.random(A.shape[0])


A = lx.MatrixLinearOperator(jnp.array(A))
M = lx.MatrixLinearOperator(jnp.linalg.inv(A.matrix), # exact inverse, should only take 1 iteration
                            tags=lx.positive_semidefinite_tag # needed bc of #139
                           )

print("======GMRES======")

for i in range(1, 7):
    print("max_steps=", i)
    out = lx.linear_solve(A, b, solver=lx.GMRES(1e-4, 1e-4, max_steps=i, restart=1), 
                              options={"preconditioner": M}, 
                              throw=False)
    x = out.value
    print(f"norm(Ax-b)={np.linalg.norm(A.mv(x)-b):.3e}")
    print("result=", lx.RESULTS[out.result]) 


print("======BiCGStab======")

for i in range(1, 7):
    print("max_steps=", i)
    out = lx.linear_solve(A, b, solver=lx.BiCGStab(1e-4, 1e-4, max_steps=i), 
                              options={"preconditioner": M}, 
                              throw=False)
    x = out.value
[...]

Issue #139, opened 2025-03-06
Clarification on docstrings for preconditioners


f0uriest:
All of the iterative methods have an option to give a preconditioner, but the docstrings are a bit confusing in a few cases:
- For GMRES, BiCGStab it says the preconditioner should be positive definite, but I don't think this is the case (and in general you probably don't want it to be PSD unless A is as well)
- For NormalCG, its unclear if the preconditioner should approximate $A^{-1}$ or $(A^TA)^{-1}$. The implementation makes it look like the latter but I think the former is more natural/common (and if it is the former then the note about the preconditioner being PSD should be removed here as well).

patrick-kidger:
I think these might be wrong! Very happy to update these if you think they should be.

jpbrodrick89:
I believe these are fixed now, no mention of psd in GMRES/BiCGStab docs and Normal correctly says

> preconditioner: A [lineax.AbstractLinearOperator](https://docs.kidger.site/lineax/api/operators/#lineax.AbstractLinearOperator) to be used as preconditioner. Defaults to [lineax.IdentityLinearOperator](https://docs.kidger.site/lineax/api/operators/#lineax.IdentityLinearOperator). This should be an approximation of the (pseudo)inverse of A. When passed to the inner solver, the preconditioner M is replaced by $MM^*$ and $M^*M$ in the first and second versions of the normal equations, respectively.

PR #162, opened 2025-06-16
Fixes for preconditioned krylov solvers

f0uriest:
- Uses transpose/conjugate preconditioner for transpose/conjugate linear solve for backwards pass. Previously the preconditioner was discarded in these cases leading to convergence issues. (#160)
- Clarifies some docstrings about the type of preconditioning used, and requirements the preconditioner must satisfy. (#139)


patrick-kidger:
Okay, awesome -- this all LGTM! (I think I see some fixes for the NormalCG case in there too :) )

Issue #149, opened 2025-05-06
Is there a way to auto-recognise operator structure?


jpbrodrick89:
I am looking to extend the `diffrax` example on the [1D heat equation](https://docs.kidger.site/diffrax/examples/nonlinear_heat_pde/) to use a Tridiagonal solve, unfortunately `lineax` can not tell a generic 2D array has tridiagonal structure. If this is not possible, could we make a change to `optimistix` to create the desired `LinearOperator` from the `jacobian`

MWE for context:
```python
import jax.numpy as jnp
import jax
import lineax
import optimistix
jax.config.update("jax_enable_x64", True)

def laplacian_numerator(y):
    return jnp.diff(jnp.diff(y), prepend=0., append=0.)

def diffus_step(y, args):
    return y + 0.1 * laplacian_numerator(y)

solver = optimistix.Newton(rtol=1e-8, atol=1e-8, linear_solver=lineax.Tridiagonal())

key = jax.random.PRNGKey(0)
rhs = jax.random.normal(key, (100,))
sol = optimistix.root_find(lambda y, args: diffus_step(y, args) - rhs, solver, rhs, max_steps=1, throw=False)
```
Returns the follow traceback
```
...
--> 220 return iterative_solve(
    221     fn,
    222     solver,
    223     y0,
    224     args,
    225     options,
    226     max_steps=max_steps,
    227     adjoint=adjoint,
    228     throw=throw,
    229     tags=tags,
    230     f_struct=f_struct,
    231     aux_struct=aux_struct,
    232     rewrite_fn=_rewrite_fn,
...
     53         "matrices"
     54     )
     55 return tridiagonal(operator), pack_structures(operator)

ValueError: `Tridiagonal` may only be used for linear solves with tridiagonal matrices
```

jpbrodrick89:
Thanks Patrick, this worked great, perfectly happy with a "promise" rather than a check. We are currently doing manual operator splitting with implicit/explicit Euler steps, so using `optimistix` directly should be fine for now. However, if we want to use higher order methods in the future, exposing the `tags` argument is probably a good idea. However, the `tags` could probably be better documented in `optimistix` perhaps by way of an example.

I can confirm that simply adding the tag in decreases runtime by about 30% on CPU on my `lineax` branch with `unroll=1` in the solver (apologies for the delay on the PR for this I am looking to see if we can improve `jax` batched performance first to greatly simplify the whole affair).

[...]

PR #164, opened 2025-06-17
Support "diagonal" primitives with no/slow JVP batch rule

jpbrodrick89:
BREAKING CHANGE: Diagonals are no longer "extracted" from operators but rely on the promise of the `diagonal` tag being accurate and fulfilled. If this promise is broken solutions may differ from user expectations.

### Preface

At present, both `JacobianLinearOperator` and `FunctionLinearOperator` require full materialisation even if provided with a `diagonal` tag. This seems self-evidently expensive (in practice it certainly can be but more often is not, see below) and requires the underlying function (which could potentially be a custom primitive) to have a batching rule. As it is currently the case that tags are considered to be a "promise" and are unchecked with no guarantee of behaviour, there are some shortcuts we can take.

### Changes made

The proposal here is to use the observation that the diagonal of a matrix can be obtained by pre/post-multiplying it by a unit vector and thereby re-write the single-dispatch `diagonal` method for `JacobianLinearOperator` and `FunctionLinearOperator` so that the `as_matrix()` method is not required. For `JacobianLinearOperator` either `jax.jvp` or `jax.vjp` will be called depending on the `jac` keyword (forward-mode should always be more efficient but meeting the user's expectation will avoid issues if forward-mode is not supported such as when using a `custom_vjp` is used). For `FunctionLinearOperator`, we can just use `self.mv`.

However, if the matrix is not actually diagonal this identity will not hold and results may be unexpected due to contributions from off-diagonals.

I considered using `operator.transpose().mv` instead of writing out `vjp` but if the matrix is tagged as `symmetric` then this would end up calling `jacrev` instead of `vjp`.

### Why is this helpful?

When using `lineax` directly one can of course just define a `DiagonalOperator` instead of a more general `JacobianLinearOperator`, but this is not always possible. For example, when using `optimistix`, the operator is instantiated within the optimisation routine and the only way to inform the optimiser about the underlying structure of the matrix is through `tags`. Therefore, if the function being optimised is a primitive (e.g. an FFI) with a JVP rule that does not support batching a user is stuck. If a slow batching rule, such as `vmap_method="sequential"`, is used the current approach is also painfully slow for large matrix sizes.

### Performance impact

I had initially hoped this to have a minor positive impact on performance across the board, but as ever I have massively underestimated the power of XLA. In practice, whether this PR seems to improve performance (e.g. for a `linear_solve` or an `optimistix.root_find`) of a _pure jax function_ appears to fluctuate with array size. By playing around with different `XLA_FLAGS` and other environment variables, my best guess is that this is mostly due to threading; a `vmap` applied to a `jnp.eye` is threaded much more aggressively meaning that the apparent time complexity appears to be of lower order than a more direct approach. However, when I tried to eliminate threading this PR still seems to have an 8–10% negative impact on performance for array sizes > 100 on an `optimistix.root_find`.

<details>
<summary>
Pure `jax` comparison: using `jvp` when attempting to enfore single-threadedness is about 14µs faster.
</summary>

```python
@jax.jit
def from_eye(x):
    eye = jnp.eye(len(x), dtype=x.dtype)
    jac = jax.vmap(lambda t: jnp.cos(x)*t, out_axes=-1)(eye)
    return jnp.diag(jac)

@jax.jit
def direct(x):
    return jnp.cos(x)
```
It seems self-evident that the second function should be more efficient, however with the new `thunk` runtime on my Mac `from_eye` runs faster than `direct` (referred to as `wrapped` in the diagram below, you can ignore `unwrapped` and `vmap` as similar performance) for array sizes > ~1.5E4:
image

Disabling the `thunk` runtime (with `XLA_FLAGS=--xla_cpu_use_thunk_runtime=false` which is reported to run faster in some circumstances) decreases the gap between the two by slightly slowing down the `eye` implementation and accelerating the `direct` approach:
image

Going further and following all suggestions in [github.com/jax-ml/jax/discussions/22739](https://github.com/jax-ml/jax/discussions/22739) to limit to one thread/core and we can see the `direct` approach is now consistently about 14µs faster:
image
</details>

<details>
<summary>
`linear_solve` significantly faster (often >2x) for array sizes <2E4 using thunk runtime, but runs about 6–10% slower for large array sizes when disabling and attempting to enforce a single thread
</summary>

<details>
<summary>
Code tested
</summary>

```python
solver = lineax.Diagonal(well_posed=True)

def double(x, args):
    return (2.0 * jnp.ones_like(x)) * x

@jax.jit
def jac_op_to_sol(rhs):
    op = lineax.JacobianLinearOperator(double, rhs, tags=frozenset({lineax.diagonal_tag}))
    return lineax.linear_solve(op, rhs, solver, throw=False)

@jax.jit
def func_op_to_sol(rhs):
    op = lineax.FunctionLinearOperator(lambda x: double(x, None), jax.eval_shape(lambda: rhs), tags=frozenset({lineax.diagonal_tag}))
    return lineax.linear_solve(op, rhs, solver, throw=False)
```
</details>

Using standard thunk runtime and `EQX_ON_ERROR=nan` we see significant speedup for array sizes < 1E4
image

Enforcing single-thread the performance between the old and the new approaches is very similar but tracks at about 6–10% slower for larger array sizes.
image

(Note that `DiagonalOperator` is actually slower somehow.)

</details>

<details>
<summary>
Similar behaviour is observed with `optimistix.root_find` (but with more modest gains, and some hits for larger array sizes)
</summary>

I compared performance for a multi-root find of the `sin` function (with `EQX_ON_ERROR=nan`):
```python
x = jnp.repeat(jnp.linspace(-10., 10., num=20), n // 20)
newton = optx.Newton(rtol=1e-8, atol=1e-8, linear_solver=lx.Diagonal())
optx.root_find(lambda x, args: jnp.sin(x), newton, x, max_steps=8, tags=frozenset({lx.diagonal_tag}), throw=False).value
```

Default settings (`jax` 0.6.1, `main` vs this branch of `lineax) with and without standard `thunk` runtime:
image

In both runtimes this PR improves/maintains performance by up to a factor of 2 for arrays of size up to 1E4 at which point it becomes slightly slow than the current version (by ~8%).

However, limiting to one thread as best as I can most of the noise is eliminated and the two have very similar performance time (the change tracking about 6% slower) except for an array size of 20 where the proposed change is faster:
image

</details>

<details>
<summary>
Much more substantial performance improvement (8x or higher) is observed for primitives that only support `sequential` batching rules
</summary>

This is a very contrived example, but based on very real use cases we have over at [tesseract-core](https://github.com/pasteurlabs/tesseract-core) and [tesseract-jax](https://github.com/pasteurlabs/tesseract-jax). I have defined a new primitive version of `sin` with a `jvp` rule that batches sequentially and is therefore slow and doesn't benefit from compilation/threading in the same way:

<details>
<summary>
Code for primitives
</summary>

```python
import numpy as onp

import jax
import jax.numpy as jnp
from jax.interpreters import ad, batching, mlir
from jax.core import ShapedArray
from jax.extend import core

from jax._src.lib.mlir.dialects import hlo
from jax._src.ffi import ffi_batching_rule

from functools import partial

sin_p = core.Primitive("mysin")
cos_mult_p = core.Primitive("cos_mult")

def sin_prim(x):
    return sin_p.bind(x)

def cos_mult_prim(*args):
    p = args[0]
    return cos_mult_p.bind(*args,
                           vmap_method="sequential",
                           result_avals=(ShapedArray(p.shape, p.dtype), ))

def cos_mult_impl(*args, vmap_method="sequential", result_avals=None):
    del vmap_method, result_avals
    return onp.cos(args[0]) * args[1]

def cos_mult_abstract_eval(*args, vmap_method="sequential", result_avals=None):
    del vmap_method, result_avals
    return ShapedArray(args[0].shape, args[0].dtype)

def sin_jvp(p, t):
    return sin_prim(p[0]), cos_mult_prim(p[0], t[0])

def sin_lowering(ctx, x):
    return [ hlo.SineOp(x).result ]

def cos_mult_lowering(ctx, p, t, *, vmap_method="sequential", result_avals=None):
    del vmap_method, result_avals
    return [ hlo.MulOp(hlo.CosineOp(p), t).result ]

def cos_mult_batch(args, axes, vmap_method="sequential", result_avals=None):
    as_tup, out_dims =  ffi_batching_rule(cos_mult_p, args, axes, 
                            vmap_method="sequential",
                            result_avals=(ShapedArray(args[0].shape, args[0].dtype),))
    return jnp.stack(list(as_tup)), out_dims[0]

sin_p.def_impl(onp.sin)
cos_mult_p.def_impl(onp.cos)
sin_p.def_abstract_eval(lambda x: ShapedArray(x.shape, x.dtype))
cos_mult_p.def_abstract_eval(cos_mult_abstract_eval)
mlir.register_lowering(sin_p, sin_lowering)
mlir.register_lowering(cos_mult_p, cos_mult_lowering)
ad.primitive_jvps[sin_p] = sin_jvp
# If the `cos_mult_p` batching rule isn't present this won't run on the current version
# but will with this PR
batching.primitive_batchers[cos_mult_p] = cos_mult_batch
```
</details>

I then ran the same tests as before but with `sin_p` instead of `jnp.sin` and we can see the time complexity of the current version is almost quadratic for array sizes greater than 100 (as one would naively expect for a dense jacobian) meaning that speedups range from a factor of 2 (array size of 20) to a factor of  8 (array size of 5000) and higher:
image

</details>
 
~~Running `benchmarks/solver_speed.py` shows a negligible improvement in the single `Diagonal` solve but a 50% faster batch solve, this could of course be down to noise as the solve is only timed once.~~ (This uses `lx.Diagonal` so not relevant and probably just a fluke.)

### Testing done
- [x] CI passes after modifying `test_diagonal` such that operators are actually initialised with diagonal matrices
- [x] Can find root (using both `Newton` and `Bisection`) of scalar function with no batching rule and take gradients through the root solve (not possible previously) this tests both `JacobianLinearOperator` and `FunctionLinearOperator` in action
- [x] Can obtain `diagonal` from `JacobianLinearOperator` with `jac="bwd"` 

Happy to perform any further requested testing you see fit/necessary. I appreciate I haven't managed to test reverse-mode especially extensively.

### Next steps

In a future PR, I would like to do something similar for other structures (e.g. tridiagonal) this should address the large O(n) discrepancy observed in #149 (but not the O(0) discrepancy). I believe this will be a much more consistent and meaningful gain than observed here. This PR here should likely be a lot easier to grok and reason about the concept and discuss framework/design choices (although maybe not the performance impact :) ) before building out further.


lineax/_operator.py
```diff
@@ -1363,10 +1363,40 @@ def diagonal(operator: AbstractLinearOperator) -> Shaped[Array, " size"]:
 
 @diagonal.register(MatrixLinearOperator)
 @diagonal.register(PyTreeLinearOperator)
```
patrick-kidger, line 1365: Side note, I think we could make the `PyTreeLinearOperator` case a little more efficient by calling `jnp.diag` on each 'diagonal leaf', and concatenate those together. (You don't have to do that here, just spotting it.)

lineax/_operator.py
```diff
@@ -1363,10 +1363,40 @@ def diagonal(operator: AbstractLinearOperator) -> Shaped[Array, " size"]:
...
+    return jnp.diag(operator.as_matrix())
+
+
 @diagonal.register(JacobianLinearOperator)
+def _(operator):
+    with jax.ensure_compile_time_eval():
+        flat, unravel = strip_weak_dtype(
+            eqx.filter_eval_shape(jfu.ravel_pytree, operator.in_structure())
+        )
+        basis = jnp.ones(flat.size, dtype=flat.dtype)
```
patrick-kidger, line 1376: Could this just be `basis = jtu.tree_map(jnp.ones_like, operator.in_structure())`?

Either way, why is the compile-time-eval needed here?

lineax/_operator.py
```diff
@@ -1363,10 +1363,40 @@ def diagonal(operator: AbstractLinearOperator) -> Shaped[Array, " size"]:
...
+
+    if operator.jac == "fwd" or operator.jac is None:
+        diag_as_pytree = operator.mv(unravel(basis))
+    elif operator.jac == "bwd":
+        fn = _NoAuxOut(_NoAuxIn(operator.fn, operator.args))
+        _, vjp_fun = jax.vjp(fn, operator.x)
+        diag_as_pytree = vjp_fun(unravel(basis))
```
patrick-kidger, line 1383: I think this will fail for operators with different input and output structures. (They might still be mathematically square, just split up into a pytree in different ways / with different dtypes.) This needs to be a basis formed from `operator.out_structure()`.

patrick-kidger (commented):
Nice, I really like seeing a colouring approach like this!

I will note however that `lx.diagonal` is documented as `Extracts the diagonal from a linear operator, and returns a vector`, which is meant to include extracting the diagonal from nondiagonal operators. I think the implementations you have here should check with `is_diagonal` to determine whether to dispatch to the new or the old implementation.

patrick-kidger:
> Happy to address these during the week, just wanted to check: I did a quick trawl through the code and I don't think diagonal or tridagonal is ever used by operators that don't have the corresponding tag, so we have an option to just change the doc string instead of you like?

However they're actually a standalone public API themselves, which could be used independent of the solvers. :)

lineax/_operator.py
jpbrodrick89: I used your suggestion in my original commit but this seems to run faster based on my testing, I think both compile time evaluation (which was inspired by sparsejac implementation) and a single ones call led to performance improvements but I can try and dig in/disambiguate a bit if interest.

patrick-kidger: Hmm, curious. I could see the compile-time-eval tricking XLA into constant-propagating where it might otherwise choose not to. And perhaps having a single constant rather than n constants also leads to some kind of performance improvement somehow.

Happy to leave this as-is if this is apparently fastest.

lineax/_operator.py
jpbrodrick89: I just added a test for mismatched input and output structure and looks like these are not allowed to be diagonal in the current version of lineax. Shall we leave relaxing that assumption and fixing this for future work? This implementation should work fine as long as the assumption holds.

```
FAILED tests/test_operator.py::test_is_symmetric[float64] - ValueError: Symmetric matrices must have matching input and output structures. Got input structure (ShapeDtypeStruct(shape=(2,), dtype=float64), ShapeDtypeStruct(shape=(), dtype=float64)) and output structure (ShapeDtypeStruct(shape=(), dty...
FAILED tests/test_operator.py::test_is_symmetric[complex128] - ValueError: Symmetric matrices must have matching input and output structures. Got input structure (ShapeDtypeStruct(shape=(2,), dtype=complex128), ShapeDtypeStruct(shape=(), dtype=complex128)) and output structure (ShapeDtypeStruct(shape=(...
FAILED tests/test_operator.py::test_is_diagonal[float64] - ValueError: Symmetric matrices must have matching input and output structures. Got input structure (ShapeDtypeStruct(shape=(2,), dtype=float64), ShapeDtypeStruct(shape=(), dtype=float64)) and output structure (ShapeDtypeStruct(shape=(), dty...
FAILED tests/test_operator.py::test_is_diagonal[complex128] - ValueError: Symmetric matrices must have matching input and output structures. Got input structure (ShapeDtypeStruct(shape=(2,), dtype=complex128), ShapeDtypeStruct(shape=(), dtype=complex128)) and output structure (ShapeDtypeStruct(shape=(...
```

jpbrodrick89: Never mind, these are `PyTreeLinearOperators` but I need to address for `JacobianLinearOperators`

jpbrodrick89:
Sorry for abandoning this for so long as the day job took over. I returned as it was found to provide orders of magnitude impact on a problem I was working (root finding over multiple interpolations) even for small array sizes (200–2000). I have addressed the main point of retaining extraction of the diagonal when the diagonal tag is missing. However, your other two comments seem at odds with each other: either we ensure input/output structures match (which is actually the case, see above) enabling us to extract diagonal leafs for `PyTreeLinearOperator` and leaving the `JacobianLinearOperator` vjp implementation as it is, or we relax this assumption and are no longer able to extract diagonal leafs as there is not guaranteed to be such an object.

patrick-kidger:
Sorry for the long delay getting back to you, some personal life things took over for a while.

So, now to actually answer your question: good point.

I imagine we could probably do the 'diagonal leaf' approach when the structures match, and go for the more expensive approach when they don't?

jpbrodrick89:
Note this is largely ready to go and includes #195 (but not #196), happy to get those in first and merge this in after for a cleaner diff.

Main change I've made is the `try_sparse_materialise` helper which converts an operator to `DiagonalLinearOperator` if it's diagonal, which can be used in `mv` or `materialise` and should make the code more maintainable and extensible (I shouldn't need to touch materialise in the tridiagonal PR except for JacobianLinearOperator. This is a bit of a breaking change though as materialise does not just return self for as many operators but instead materialises as the "most efficient" representation of the operator. Happy to hear any thoughts. :-)

Also happy to be targeting a dev branch if you prefer as these are not just "fixes", one just doesn't exist right now.

lineax/_operator.py
```diff
@@ -329,6 +332,14 @@ def __init__(self, value):
         self.value = value
 
 
+def _leaf_from_keypath(pytree: PyTree, keypath: jtu.KeyPath) -> Array:
```
patrick-kidger: Nit, can we move this function closer to its callsite (much further down in the file)?

patrick-kidger (commented):
Okay indeed final thoughts – I think this all looks pretty good to me! Minor comments below only.

Whilst I think this is fine and I am happy to merge this as-is, I am wondering if the many special-cases are indicative of the need for some other structure here. (E.g. maybe we should arrange for `materialise` to always start with `_try_sparse_materialise`, and then fall back to per-operator overloads?) In general I'm not sure what that is though.

I don't think I'd like to see too much more complexity here beyond what we have now, though, e.g. I'm a bit antsy about what the analogous tridigonal implementations will look like layered on top of this.

lineax/_operator.py
```diff
@@ -1917,17 +2008,20 @@ def _(operator):
...
+    maybe_sparse_op = _try_sparse_materialise(operator)
+    if maybe_sparse_op is not operator:
+        return maybe_sparse_op
+    return materialise(operator.operator1) @ materialise(operator.operator2)
 
 
 @diagonal.register(ComposedLinearOperator)
 def _(operator):
+    if is_diagonal(operator):
+        return diagonal(operator.operator1) * diagonal(operator.operator2)
```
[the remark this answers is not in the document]
jpbrodrick89: Oh wait you mean the converse might not be true, yes good catch theoretically you're right. But right now in lineax that's not true as we don't look for tags on the composed linear operator. So if it's not made of two diagonal operator it won't REGISTER as being diagonal so it's fine for now unless you want to change that behaviour.

jpbrodrick89:
Tests still pass locally I'm wondering if its due to RNG (i.e. random vec is sometimes in range of singular matrix and sometimes not)?

Issue #175, opened 2025-11-12
LSMR init overflow


GJBoth:
Thanks for implementing the LSMR solver - I've been using it for matrix-free Levenberg-Marquardt and it seems to work really well! There's one issue with the initialisation however - [one of the initial values gets initialised as `1e100` leading to overflows.](https://github.com/patrick-kidger/lineax/blob/6bf1a0e17c6352b96bdf52e72772fac63eb641df/lineax/_solver/lsmr.py#L181). The origin of this value seems to be the [scipy implementation](https://github.com/patrick-kidger/lineax/blob/6bf1a0e17c6352b96bdf52e72772fac63eb641df/lineax/_solver/lsmr.py#L181), which runs in 64bit by default - but Jax runs in 32. It doesn't seem to affect the result of the solver, though.

Simply changing this line to:
``` 
minrbar=jnp.inf
``` 

solves the issue for me.

johannahaffner:
This just reminded me of this issue: https://github.com/patrick-kidger/optimistix/pull/186 - let's try to avoid using a `jnp.something` default value. How about we set `minrbar` to the highest value that does not overflow in 32-bit, so `1e38` if this is a float, and `1e9` if it is an integer.

PTNobel:
I think any big number should be fine ...

Issue #172, opened 2025-08-31
Test failure with 0.7.1: condition number too high for LSMR solver; test passes on 0.7.0


johannahaffner:
I ran the Lineax test suite for 0.7.1, with Equinox from its current `main` branch. 

We now get an `EquinoxRuntimeError` for `LSMR`. 

```
tests/test_vmap_jvp.py:129: EquinoxRuntimeError
=========================== short test summary info ============================
FAILED tests/test_vmap_jvp.py::test_vmap_jvp[float64-construct_singular_matrix-True-make_matrix_operator-solver20-tags20-True] - EquinoxRuntimeError: Above is the stack outside of JIT. Below is the stack ...
```

The Lineax error message is 

```
equinox.EquinoxRuntimeError: Condition number of A seems to be larger than `conlim`.
```

This test passes on 0.7.0.
I have not yet investigated this error, just opening this issue so it can be tracked.

johannahaffner:
> so it's possible we just got unlucky with the randomly generated matrix?

I think so too.

PR #176, opened 2025-11-14
Decrease default value to prevent overflow in 32-bit.

johannahaffner:
Fixes https://github.com/patrick-kidger/lineax/issues/175#issuecomment-3530551020.

### DIFF
--- a/lineax/_solver/lsmr.py
+++ b/lineax/_solver/lsmr.py
@@ -42,6 +42,7 @@ import jax.tree_util as jtu
 from equinox.internal import ω
 from jaxtyping import Array, PyTree
 
+from .._misc import complex_to_real_dtype
 from .._norm import two_norm
 from .._operator import AbstractLinearOperator, conj
 from .._solution import RESULTS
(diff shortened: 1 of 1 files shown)


patrick-kidger:
Should this be something like `jnp.finfo(jnp.float32).max`? If nothing else to document the choice here.

johannahaffner:
Sure, I can add a comment! A `jnp` anything in a default value of a public function causes JAX to initialise upon import of the library, I'm not exactly sure if this would be triggered here as well and will check if that would be the case.

johannahaffner:
We're getting the test failure reported here, which seems to be due to numerical bad luck: https://github.com/patrick-kidger/lineax/issues/172

johannahaffner:
...and another one in GMRES, although I don't see how that one could be due to the changes I made.

tests/helpers.py
```diff
@@ -59,6 +59,17 @@ def _construct_matrix_impl(getkey, cond_cutoff, tags, size, dtype, i):
...
+    # condition numbers (as they are expected to do). The way this plays out can depend
+    # on the JAX version and on the platform the code runs - with a given seed and some
+    # new and optimised XLA kernel fusion, a test may fail.
+    # The cutoffs below are a little empirical, I've chosen them to be as high as
+    # possible to ensure that we are still testing a real scenario, without generating
+    # operators that the iterative solvers aren't built for.
+    elif isinstance(solver, lx.GMRES):
+        cond_cutoff = 900
+    elif isinstance(solver, lx.LSMR):
+        cond_cutoff = 800
```
johannahaffner: This is the relevant change - I'm changing the maximum condition number of the operator for LSMR, since it will fail tests if the cutoff is at `1e3`. If this isn't lowered, then tests such as the `vmap_jvp` will fail. 

I'm not familiar enough with LSMR to understand how the condition number of the original operator propagates to the condition number we're dealing with when computing the JVP, which is another linear solve. 

FWIW we've seen other programs that were at or near their numerical limit fail when a new JAX version drops - and work again in a subsequent one. Optimisations in how XLA Fuses operations can mean that numerical errata get propagated a little differently from version to version, and if the program already operators near instability this can put it over the edge. That said, I'm not sure that LSMR *should* be at its numerical limits for the JVP of a linear solve where the original operator has a condition number of `1e3`.

Issue #24, opened 2023-06-07
Linear solvers for sparse matrices


arpastrana:
This is just a question: are there any plans to support JAX-compatible linear solvers for sparse matrices? I am thinking of sparse linear systems of the type $Ax=b$ where $A$ is sparse.

patrick-kidger:
Right! So this is doable like so:
```python
import equinox as eqx
import lineax as lx
import jax.experimental.sparse as js
import jax.numpy as jnp

class SparseMatrixLinearOperator(lx.MatrixLinearOperator):
    def mv(self, vector):
        return js.sparsify(lambda m, v: m @ v)(self.matrix, vector)

x = jnp.array([[1.0, 0.0], [0.0, 1.0]])
x = js.BCOO.fromdense(x)
op = SparseMatrixLinearOperator(x)
vec = jnp.array([1., 2.])

@eqx.filter_jit
def f(op, vec):
    sol = lx.linear_solve(op, vec, solver=lx.GMRES(rtol=1e-5, atol=1e-5))
    return sol

print(f(op, vec).value)
```
basically just (a) overriding the `mv` (matrix-vector product) computation to work with a sparse matrix, and (b) using an iterative solver (here GMRES) rather than something like LU (which requires materialising the matrix).

This isn't built into Lineax by default as JAX's sparse support is still pretty experimental, and we want to see how that's going to pan out before we design an API around it.

DoTulip:
Hello, @patrick-kidger.

I am a new user of lineax, I would like to ask if lineax supports sparse matrix AD now?

patrick-kidger:
Only to whatever extent JAX itself does.

DoTulip:
Thank you very much for your reply!

patrick-kidger:
Oh yikes, that's not a fun one. My guess is that this is a bug in JAX itself -- its sparse implementation does some fairly complicated things in its tracing mechanisms.

(CC @johnviljoen here who I believe is interested in sparsity.)

I'm sorry that I can't be more help. If you determine a fix, whether in JAX or in Lineax, then I'd be curious to know what you find.

johannahaffner:
Thank you! Canvassing insights like these is exactly what I hoped to do here :) 

> The BCOO etc pytrees instead try to record their batched-ness, and hook into the vmap machinery to do so.
> 
[...]

PR #178, opened 2025-11-24
Sparse operators in Lineax: CSR/CSC and COO format

johannahaffner:
These operators wrap the respective JAX operators. They are currently limited to 2D matrices, which is what JAX supports as well. As discussed in https://github.com/patrick-kidger/lineax/issues/24#issuecomment-3571451568, this introduces a dependency on `jax.experimental.sparse`, which while part of the experimental module does seem to be actively maintained. This is mostly necessary so that we can use their implementation of `__matmul__`. 

It turns out I had the COO version already, I had forgotten about that 😃 

As a drive-by change, this adds extra fixtures for cache clearing, which were necessary to get tests to pass on the Linux machine I was using while coding this up. 

Some loose strings to tie up: 

- When not type-checking, then the batched sparse format (`BCOO`, `BCSR`) may be substituted for their bare-bones, non-batched counterparts. I have not checked whether these pass all the tests, and whether their special batching rules would interfere with what we're already doing. Vmapping works without using these formats.
- I'll expand the documentation of both, to also include the limitation to 2D matrices explicitly and I'll add an example for their usage.

### DIFF
--- a/docs/api/operators.md
+++ b/docs/api/operators.md
@@ -74,3 +74,42 @@ Or, perhaps we only have a function $F : \mathbb{R}^m \to \mathbb{R}^n$ such tha
     options:
         members:
             - __init__
+
+---
+
+## Sparse Linear Operators
+
+Lineax supports two flavours of sparse representations of linear operators by wrapping implementations from the `jax.experimental.sparse` module. Representations of sparse operators can be created from a dense matrix, or directly from values and row and column indices in a given format. In either case, Lineax only ever sees the existing JAX representation of the sparse operator.
+Note that the API of `jax.experimental.sparse` is still subject to change, and the API offered here may therefore change as well.
+
+!!! Example
+    
+    ```python
+    import lineax as lx
+    from jax.experimental.sparse import CSR
+
+    matrix = ...  # Some 2D matrix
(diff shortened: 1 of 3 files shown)


johannahaffner:
Yes, that is pretty much where I land on this as well. I don't think this should be part of the next release, at the very least a lot of sparsity-specific testing needs to be done first - it could be that we circumvent a lot of edge cases by virtue of testing these with solvers that never actually need to interact with the sparse structures themselves, and either work with the densely materialised matrix or matrix-vector products.

Not relying on `jsp.COO` etc. at all requires implementing our own `__matmul__` for these structures, and that is a bit more involved. The BCSR formats also have other special features to sum operators with the same shape but different sparsity structures - these are useful when you need to add a diagonal for example, for regularisation. These are all quite good, and rewriting this would be a big technical task. 

FWIW I would not call `jax.experimental.sparse` unmaintained, it is in maintenance mode but not actively developed any further. It does seem to be quite stable and has been around for quite a while now. 

TLDR - let's not merge this now, leave it here and / or revisit it for BCSR later. Right now the only use case is CuDSS anyway, and the operator is basically a vessel to communicate to CuDSS what the values, offsets and columns are. We can do all of this in Spineax.

johannahaffner:
Closing this for now.

PR #179, opened 2025-11-24
Update guidelines for installation of doc dependencies

johannahaffner:
The contribution guidelines were out of date since the doc stack update. 

This PR is expected to fail the CI, with the same things that fail in https://github.com/patrick-kidger/lineax/pull/176 and and https://github.com/patrick-kidger/lineax/pull/178.


patrick-kidger:
I'm going to close this as I'm about to update the rest of our infra in the same way as I've done for the other projects, and do this at the same time :)

PR #184, opened 2025-12-27
Replace deprecated implicit conversion of an array to a dtype in `MatrixLinearOperator` initialization

RunnersNum40:
`MatrixLinearOperator` initialization uses an implicit dtype conversion to check if the matrix has an inexact type. JAX in deprecated this in [0.4.30](https://docs.jax.dev/en/latest/changelog.html#jax-0-4-30-june-18-2024) and it will be an error in [0.8.0](https://docs.jax.dev/en/latest/changelog.html#jax-0-8-0-october-15-2025).

This PR replaces it with an explicit check.

Here's an MVP of the issue:

```py
import warnings

import jax.numpy as jnp
import lineax as lx


warnings.simplefilter("default", DeprecationWarning)


lx.MatrixLinearOperator(jnp.eye(1))
```

And the result is:
`DeprecationWarning: Implicit conversion of an array to a dtype is deprecated; rather than dtype=arr use dtype=arr.dtype. In the future this will result in an error.`

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -10,14 +10,14 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Release
-        uses: patrick-kidger/action_update_python_project@v6
+        uses: patrick-kidger/action_update_python_project@v8
         with:
             python-version: "3.11"
             test-script: |
                 cp -r ${{ github.workspace }}/tests ./tests
                 cp ${{ github.workspace }}/pyproject.toml ./pyproject.toml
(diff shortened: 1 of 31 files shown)


RunnersNum40:
Dang, I thought I looked hard enough for other cases. There's at least nothing else that matches theses two instances.

After pointing at `dev` I realized that both of these were fixed already in https://github.com/patrick-kidger/lineax/commit/9c1b9a9f3e2a3ae55235a996fec357444d8a7d95. I'm closing this PR in light of that.

PR #191, opened 2026-01-28
fix linearise for JacobianLinearOperator with jac=bwd and use linear_transpose in mv

jpbrodrick89:
It turns out `lx.linearise` fails for JacobianLinearOperator with `jac="bwd"` if there is custom vjp which prevents forward-mode autodiff. This was missed in the test because just CALLING `lx.linearise` is fine—we can create a jaxpr representing the failure, but an error is only raised on EVALUATION. I updated the tests so they would fail and corrected linearise to use jax.linear_transpose (it turns out you can use these even with a custom vjp!).

I also utilised linear_transpose in the mv so we're not computing `jacrev` each time, however for most use cases a user should ALWAYS want to call `lx.linearise` on such a JacobianLinearOperator to avoid recomputing the primal (unlike using jvp with fwd mode, linear_transpose has no memory advantages so there is no reason to not cache it for reuse unless you know you're only going to need a single `mv`).

Users who used custom vjp's that are nonlinear or affine in their cotangents will now get jax errors, this is expected and such custom vjp's are fundamentally incorrect. I had to correct one of the tests that used an affine custom vjp to use a linear one.

These improvements should simplify the coloring method PR's I have open for JacobianLinearOperator as `operator.T.mv` will now work with custom vjp's if symmetric (previously this failed and we had to right out the backward mode more verbosely).

### DIFF
--- a/tests/test_operator.py
+++ b/tests/test_operator.py
@@ -300,7 +300,12 @@ def test_is_tridiagonal(dtype, getkey):
 @pytest.mark.parametrize("dtype", (jnp.float64, jnp.complex128))
 def test_tangent_as_matrix(dtype, getkey):
     def _list_setup(matrix):
-        return list(_setup(getkey, matrix))
+        # Exclude jacrev operator: jac="bwd" uses custom_vjp which doesn't support JVP
+        return [
+            op
+            for op in _setup(getkey, matrix)
+            if not (isinstance(op, lx.JacobianLinearOperator) and op.jac == "bwd")
+        ]
 
     matrix = jr.normal(getkey(), (3, 3), dtype=dtype)
     t_matrix = jr.normal(getkey(), (3, 3), dtype=dtype)


lineax/_operator.py
```diff
@@ -543,15 +543,20 @@ class JacobianLinearOperator(AbstractLinearOperator):
...
+    forward-mode differentiation.
 
     See also [`lineax.materialise`][], which materialises the whole Jacobian in
     memory.
+
+    !!! tip
+
+        For repeated `mv()` calls, consider using [`lineax.linearise`][] to cache
+        the primal computation. This is especially beneficial with `jac="bwd"`
+        as the primal computation affects the entire backward pass.
```
patrick-kidger: Is there a difference between fwd/bwd in efficiency here? In either case it's about computing the primal pass and saving the residuals needed for the subsequent tangent/cotangent sweep.

lineax/_operator.py
```diff
@@ -1238,10 +1251,33 @@ def _(operator):
...
+            )
+        else:
+            # Transpose the VJP to get J @ v from J.T @ v
+            transpose_vjp = jax.linear_transpose(
+                lambda g: vjp_fn(g)[0], operator.out_structure()
+            )
+
+            def mv_fn(v):
+                (out,) = transpose_vjp(v)
+                return out
```
patrick-kidger: This is `_Unwrap` I think

patrick-kidger:
Awesome, merged! 🎉

And on `test_tangent_as_matrix`, yup I think skipping this should be fine. As you note it's not really defined (I think regardless of whether the `custom_vjp` actually occurs at the module level).

PR #195, opened 2026-02-02
ensure materialise and linearise preserves aux

### LINKED ISSUE #194: Inconsistent behaviour of materialise wrt Aux
My understanding of materialise is that after calling it a first time, calling it again should have no impact. However, JacobianLinearOperator's materialise returns an AuxLinearOperator but AuxLinearOperator's materialise strips the aux. Which is the correct behaviour (strip or keep aux)?

jpbrodrick89:
Fixes #194 , currently `materialise(JacobianLinearOperator)` returns an `AuxLinearOperator` but `materialise(AuxLinearOperator)` returns the materialised inner operator with the aux stripped. After this PR, `materialise(AuxLinearOperator)` will preserve the aux by returning `AuxLinearOperator(linearise(operator.operator), operator.aux)`.

The same change has been made for linearise.

I also make a change to the handling of linearise/materialise for `ComposedLinearOperator` which changes it to returning an `AuxLinearOperator`, if the FIRST operator has aux. This means `materialise(...).aux` will always work if the original operator returned some kind of aux.

Note we currently have ill-defined behaviour if a user tries something strange like adding two operators that have aux or applying an operator to an aux operator. We might want to either make explicit the behaviour that we (may) silently strip aux, or catch this early and recommend the user to use an equivalent of `_NoAuxOut`.

### DIFF
--- a/lineax/_operator.py
+++ b/lineax/_operator.py
@@ -1863,9 +1863,19 @@ for transform in (linearise, materialise, diagonal):
     def _(operator, transform=transform):
         return transform(operator.operator) / operator.scalar
 
-    @transform.register(AuxLinearOperator)  # pyright: ignore
+
+# diagonal strips aux (returns array, not operator)
+@diagonal.register(AuxLinearOperator)
+def _(operator):
+    return diagonal(operator.operator)
+
+
+# linearise and materialise preserve aux
+for transform in (linearise, materialise):
+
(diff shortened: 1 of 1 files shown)


jpbrodrick89:
Yes I support removing AuxLinearOperator, it only seems to be used when materialising JacobianLinearOperator which seems very "special casing".

PR #200, opened 2026-02-03
fix is_symmetric for complex PSD/NSD matrices

jpbrodrick89:
Currently, is_symmetric returns True for PSD/NSD matrices. This led to failing test in test_well_posed.py when we changed our transpose rules, this should be a minimal fix.

Note we could go all out and add `is_hermitian` tests and `.H` helper function for less verbosity, but as I think we're not really prioritising complex support right now (e..g see comment in CG solver) that can be delayed for now.

### DIFF
--- a/lineax/_operator.py
+++ b/lineax/_operator.py
@@ -1782,7 +1782,7 @@ def _(operator):
 
 @is_positive_semidefinite.register(IdentityLinearOperator)
 def _(operator):
-    return True
+    return eqx.tree_equal(operator.in_structure(), operator.out_structure()) is True
 
 
 @is_positive_semidefinite.register(DiagonalLinearOperator)


patrick-kidger (commented):
One comment only, LGTM!

patrick-kidger:
Okay, deciding to go ahead and merge as this is blocking other stuff I'm working on :D (I'll incorporate my comment into #203.)

jpbrodrick89:
Ah, sorry, i actually addressed your comment and committed but I must have tried to push to your repo rather than my fork and didn't notice the error message. 🤦‍♂️ Will rebar my other PRs on yours when that's merged. Thanks!

jpbrodrick89:
Can share my version when i get back to my desk in a few hours if you haven't done it already.

PR #204, opened 2026-02-10
added LSMR to docs

### LINKED ISSUE #193: LSMR docs not added to solvers.md
Should probably be added to the least squares solvers section with a reference in the iterative solvers but could be the other way round.

patrick-kidger:
Fixes #193.

(Also I reordered a bit of this documentation whilst I was here)

### DIFF
--- a/docs/api/solvers.md
+++ b/docs/api/solvers.md
@@ -49,9 +49,48 @@ These are capable of solving ill-posed linear problems.
         members:
             - __init__
 
-!!! info
+---
 
-    In addition to these, `lineax.Diagonal(well_posed=False)` (below) also supports ill-posed problems.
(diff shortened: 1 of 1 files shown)


jpbrodrick89:
Happy with the wording and structure here, but I think the h3 heading doesn't work as well as the original info box because the current theme doesn't distinguish much between h2 and h3 and its easy to get confused what constitutes a new section. However, I do like that your restructure puts the multi-class solvers in the toc. What do you think about try h4 instead?

H3 `### Diagonal`

image

H4 `#### Diagonal`

image

patrick-kidger:
No strong feelings! Happy with whatever you think is most readable (h3, h4, admonition, ...)

PR #212, opened 2026-03-09
prevent linear_solve calling jvp of solver.init

### LINKED ISSUE #211: Differentiation of linear_solve causes differentiation through factorization (solver state)
In writing #210, I discovered a bug in jax's autograd rule for pivoted qr. However, I should have never had the opportunity to observe this bug as lineax code should never need this jvp rule, instead defining its own `linear_solve_p` primitive with custom jvp. Here is a minimal example showing the issue
```python
import jax
import lineax as lx
from lineax._solve import AbstractLinearSolver

key1, key2, key3 = jax.random.split(jax.random.key(0),3)
m = jax.random.normal(key1, (3,3))
mt = jax.random.normal(key2, (3,3))
v = jax.random.normal(key3, (3,))

class DisallowGradWrapper(AbstractLinearSolver):
    solver: AbstractLinearSolver
    def init(self, operator, options):
        @jax.custom_jvp
        def f(operator):
            return self.solver.init(operator, options)
        @f.defjvp
        def _(*args):
            assert False # This assertion gets hit
        return f(operator)
    def compute(self, state, vector, options):
        return self.solver.compute(state, vector, options)
    def transpose(self, state, options):
        return self.solver.transpose(state, options)
    def conj(self, state, options):
        return self.solver.conj(state, options)
    def assume_full_rank(self):
        return self.solver.assume_full_rank()

f = lambda m: lx.linear_solve(lx.MatrixLinearOperator(m), v, solver=DisallowGradWrapper(lx.QR())).value
res = jax.jvp(f, (m,), (mt,))
```

This code asserts that solver.init() is never differentiated through, but the assertion is hit.

jpbrodrick89:
fixes #211, previously stop_gradient was called AFTER solver.init. This meant the jvp of solver.init was still called before the return tangent being set to zero (the primal returned by the jvp is retained). If we call stop_gradient BEFORE then the jvp is never called. This means we can use primitives without jvp's (e.g. geqp3/geqrf) or with incorrect jvp primal's (e.g. qr with `pivoting=True`).

Note that we don't need to apply `stop_gradient` to options as we do not support taking a gradient with respect to e.g. precondtioners (we just get a nondiff error).

If we're happy with this I will modify my invert PR #206 to mirror the same pattern.

### DIFF
--- a/lineax/_solve.py
+++ b/lineax/_solve.py
@@ -771,7 +771,11 @@ def linear_solve(
             stats={},
         )
     if state == sentinel:
-        state = solver.init(operator, options)
+        dynamic_operator, static_operator = eqx.partition(operator, eqx.is_array)
+        stopped_operator = eqx.combine(
+            lax.stop_gradient(dynamic_operator), static_operator
+        )
+        state = solver.init(stopped_operator, options)
 
     dynamic_state, static_state = eqx.partition(state, eqx.is_array)
     dynamic_state = lax.stop_gradient(dynamic_state)
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -381,10 +381,9 @@ def jvp_jvp_impl(
         if use_state:
 
             def linear_solve1(operator, vector):
-                state = solver.init(operator, options={})
-                state_dynamic, state_static = eqx.partition(state, eqx.is_inexact_array)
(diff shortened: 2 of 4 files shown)


patrick-kidger:
Alright, merged! Thank you for this 🎉

PR #219, opened 2026-04-30
Use ormqr for more efficient QR solve

jpbrodrick89:
Typically on CPU this leads to a pretty stable 1.5x speedup in a QR solve (a 1000 times speed up in the multiplication itself but factorisation is still a heavy part of the solve), on GPU wins are mostly for large matrices but when https://github.com/jax-ml/jax/pull/36575 is merged wins will be realised across the board (sometimes by >8x). We just adopt this as the default and hide complexity from the user.

I think this is the last change I'd like to make before the next release.

@adconner I'd appreciate a quick pass over sanity check if you have time. I'm very aware that your pivoted QR PR is still open, my current thinking is that we should focus on doing a more complete least square focussed release next. The first foundation that I think might be help is having max rank tags and tracking rather than passing this to the solver. However, if you'd find this useful and rather it released now rather than in a few months I can try help iterate with you on a temporary design (I'm currently leaning towards the `QR(pivoted=True)` scipy/numpy homage API rather than `PivotedQR`).

@patrick-kidger note this means I have to bump Jax's minimum version to 0.10.0, are you comfortable with this? Does this have any implications on what version we should call the next release.

### DIFF
--- a/lineax/_solver/qr.py
+++ b/lineax/_solver/qr.py
@@ -15,6 +15,7 @@
 from typing import Any, TypeAlias
 
 import equinox.internal as eqxi
+import jax.lax.linalg as jll
 import jax.numpy as jnp
 import jax.scipy as jsp
 from jaxtyping import Array, PyTree
@@ -58,9 +59,10 @@ class QR(AbstractLinearSolver):
         transpose = n > m
         if transpose:
             matrix = matrix.T
-        qr = jnp.linalg.qr(matrix, mode="reduced")  # pyright: ignore
+        h, taus = jnp.linalg.qr(matrix, mode="raw")  # pyright: ignore
+        a = h.mT
         packed_structures = pack_structures(operator)
-        return qr, eqxi.Static(transpose), packed_structures
+        return (a, taus), eqxi.Static(transpose), packed_structures
 
     def compute(
         self,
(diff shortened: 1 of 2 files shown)


patrick-kidger:
> note this means I have to bump Jax's minimum version to 0.10.0, are you comfortable with this? Does this have any implications on what version we should call the next release.

That's okay! I'd usually trigger the downstream Optimistix/Diffrax CI suites (just manually locally, no nice setup) on top of the new JAX+Lineax to be sure that the new JAX version doesn't break anything for them i.e. that we don't release a Lineax version that is incompatible with them.

Version number bump would just be patch, not minor.

adconner:
Looks good to me. Glad to see this added to lineax and also appreciate your work improving the GPU situation with jax-ml/jax#36575. 

I think I am in general in favor of designs which let operators express rank assumptions/knowledge in tags (in addition to other properties). In the most expressive case of this design you can track rank lower bound and rank upper bound as tags. 

I briefly discussed just a full_rank tag with @patrick-kidger in the context of #158, as for instance if this is the source of truth, then (1) jvp terms are always most efficient regardless of if a more general solver is used, (2) the assume_full_rank solver special method is no longer needed, solvers just check rank tags in their __init__ just like other tags, and (3), AutoLinearSolver could be truly auto, not needing well_defined argument.

I think we landed on essentially that this feels too different from the way this assumption is communicated in other libraries, and basically that solver selection should be considered part of the source of truth for what the operators are.  Maybe if we need rank tracking for other reasons we can consider an expanded design

jpbrodrick89:
Thanks @adconner , that's a very good point of reference I had read through before and had forgotten about. I think this is why future least squares development needs a more careful design discussion, I'll probably open an issue to discuss ideas for this in a week or so.

The other idea I had for a rank tracker is a Woodbury solver that could take an `AddLinearOperator` and detect which one is full rank and which one is a low-rank `ComposedLinearOperator`, but I'm still not sure that's a good design either. 

I'll write up some release notes tomorrow for the current state.

jpbrodrick89:
Sorry @patrick-kidger I was sloppy and merged without the diffrax and optimistix tests, I will set them up to run tomorrow and flag/handle/revert if any issues.

PR #221, opened 2026-05-07
Refactor tests for random rank deficient examples and analytic correctness in the reference jvp

adconner:
Here are a number of changes to the tests, let me know what you think.

Add a reference solver jnp_lstsq which doesn't suffer from NaN derivatives (jax issue #14868) using the implicit function theorem. This makes finite differences no longer necessary in the tests.

Rank deficient test matrices are now random (obtained by zeroing a singular value), rather than structural (obtained by just zeroing a row or column or some other rank deficient pattern). The second order derivative tests require tangent directions chosen tangent to the low rank locus, so such directions are now computed by jvp through the projection to the low rank locus instead of being structural.

The main solver parameterization now records solvers by tags and whether they support rank deficient operators and nonsquare operators, rather than combining these into a single pseudoinverse boolean. Then all valid combinations are tried in all tests according to this recorded data, removing the old need for separate nonsquare testing, and now we exercise the previously untested combination of nonsquare rank deficient operators for solvers which support it.

Furthermore, tangent matrices in jvp tests are no longer chosen to have the same properties as the primal, in particular structural tags, as this is not mathematically required. It is required only in our implementation for the cases in which the structural property has been used to accelerate operator.mv in addition to solve, which currently includes just diagonal and tridiagonal tags. If more tags become exploited like this in the future, the tested tangents can be restricted by modifying _construct_matrix_impl.

### DIFF
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -613,24 +613,39 @@ def jvp_jvp_impl(
     linear_solve2 = ft.partial(eqx.filter_jvp, linear_solve1)
     jnp_solve2 = ft.partial(eqx.filter_jvp, jnp_solve1)
 
+    jnp_primals = (matrix, vec, t_matrix, t_vec)
+
     def _make_primal_tangents(mode):
-        lx_args = ([], [], operator, t_operator, tt_operator, tt_t_operator)
-        jnp_args = ([], [], matrix, t_matrix, tt_matrix, tt_t_matrix)
-        for primals, ttangents, op, t_op, tt_op, tt_t_op in (lx_args, jnp_args):
-            if "op" in mode:
-                primals.append(op)
-                ttangents.append(tt_op)
-            if "vec" in mode:
-                primals.append(vec)
-                ttangents.append(tt_vec)
-            if "t_op" in mode:
-                primals.append(t_op)
-                ttangents.append(tt_t_op)
-            if "t_vec" in mode:
-                primals.append(t_vec)
-                ttangents.append(tt_t_vec)
-        lx_out = tuple(lx_args[0]), tuple(lx_args[1])
-        jnp_out = tuple(jnp_args[0]), tuple(jnp_args[1])
+        lx_primals = []
+        lx_tangents = []
+        jnp_tangents = []
+        if "op" in mode:
+            lx_primals.append(operator)
+            lx_tangents.append(tt_operator)
+            jnp_tangents.append(tt_matrix)
+        else:
+            jnp_tangents.append(jnp.zeros_like(tt_matrix))
+        if "vec" in mode:
+            lx_primals.append(vec)
+            lx_tangents.append(tt_vec)
+            jnp_tangents.append(tt_vec)
+        else:
+            jnp_tangents.append(jnp.zeros_like(tt_vec))
+        if "t_op" in mode:
+            lx_primals.append(t_operator)
+            lx_tangents.append(tt_t_operator)
+            jnp_tangents.append(tt_t_matrix)
+        else:
+            jnp_tangents.append(jnp.zeros_like(tt_t_matrix))
+        if "t_vec" in mode:
+            lx_primals.append(t_vec)
+            lx_tangents.append(tt_t_vec)
+            jnp_tangents.append(tt_t_vec)
+        else:
+            jnp_tangents.append(jnp.zeros_like(tt_t_vec))
+
+        lx_out = tuple(lx_primals), tuple(lx_tangents)
+        jnp_out = jnp_primals, tuple(jnp_tangents)
         return lx_out, jnp_out
 
     modes = (
@@ -648,59 +663,48 @@ def jvp_jvp_impl(
         {"vec", "t_op", "t_vec"},
         {"op", "vec", "t_op", "t_vec"},
     )
+    jnp_solve3 = ft.partial(
+        eqx.filter_jvp, lambda op, v, t_op, t_v: jnp_solve2((op, v), (t_op, t_v))
+    )
+    jnp_solve3 = eqx.filter_jit(jnp_solve3)
     for mode in modes:
         if mode == {"op"}:
             linear_solve3 = lambda op: linear_solve2((op, vec), (t_operator, t_vec))
-            jnp_solve3 = lambda mat: jnp_solve2((mat, vec), (t_matrix, t_vec))
         elif mode == {"vec"}:
             linear_solve3 = lambda v: linear_solve2((operator, v), (t_operator, t_vec))
-            jnp_solve3 = lambda v: jnp_solve2((matrix, v), (t_matrix, t_vec))
         elif mode == {"op", "vec"}:
             linear_solve3 = lambda op, v: linear_solve2((op, v), (t_operator, t_vec))
-            jnp_solve3 = lambda mat, v: jnp_solve2((mat, v), (t_matrix, t_vec))
         elif mode == {"t_op"}:
             linear_solve3 = lambda t_op: linear_solve2((operator, vec), (t_op, t_vec))
-            jnp_solve3 = lambda t_mat: jnp_solve2((matrix, vec), (t_mat, t_vec))
         elif mode == {"t_vec"}:
             linear_solve3 = lambda t_v: linear_solve2(
                 (operator, vec), (t_operator, t_v)
             )
-            jnp_solve3 = lambda t_v: jnp_solve2((matrix, vec), (t_matrix, t_v))
         elif mode == {"op", "t_op"}:
             linear_solve3 = lambda op, t_op: linear_solve2((op, vec), (t_op, t_vec))
-            jnp_solve3 = lambda mat, t_mat: jnp_solve2((mat, vec), (t_mat, t_vec))
         elif mode == {"op", "t_vec"}:
             linear_solve3 = lambda op, t_v: linear_solve2((op, vec), (t_operator, t_v))
-            jnp_solve3 = lambda mat, t_v: jnp_solve2((mat, vec), (t_matrix, t_v))
         elif mode == {"vec", "t_op"}:
             linear_solve3 = lambda v, t_op: linear_solve2((operator, v), (t_op, t_vec))
-            jnp_solve3 = lambda v, t_mat: jnp_solve2((matrix, v), (t_mat, t_vec))
         elif mode == {"vec", "t_vec"}:
             linear_solve3 = lambda v, t_v: linear_solve2(
                 (operator, v), (t_operator, t_v)
             )
-            jnp_solve3 = lambda v, t_v: jnp_solve2((matrix, v), (t_matrix, t_v))
         elif mode == {"op", "vec", "t_op"}:
             linear_solve3 = lambda op, v, t_op: linear_solve2((op, v), (t_op, t_vec))
-            jnp_solve3 = lambda mat, v, t_mat: jnp_solve2((mat, v), (t_mat, t_vec))
         elif mode == {"op", "vec", "t_vec"}:
             linear_solve3 = lambda op, v, t_v: linear_solve2((op, v), (t_operator, t_v))
-            jnp_solve3 = lambda mat, v, t_v: jnp_solve2((mat, v), (t_matrix, t_v))
         elif mode == {"vec", "t_op", "t_vec"}:
             linear_solve3 = lambda v, t_op, t_v: linear_solve2(
                 (operator, v), (t_op, t_v)
             )
-            jnp_solve3 = lambda v, t_mat, t_v: jnp_solve2((matrix, v), (t_mat, t_v))
         elif mode == {"op", "vec", "t_op", "t_vec"}:
             linear_solve3 = lambda op, v, t_op, t_v: linear_solve2((op, v), (t_op, t_v))
-            jnp_solve3 = lambda mat, v, t_mat, t_v: jnp_solve2((mat, v), (t_mat, t_v))
         else:
             assert False
 
         linear_solve3 = ft.partial(eqx.filter_jvp, linear_solve3)
         linear_solve3 = eqx.filter_jit(linear_solve3)
-        jnp_solve3 = ft.partial(eqx.filter_jvp, jnp_solve3)
-        jnp_solve3 = eqx.filter_jit(jnp_solve3)
 
         (primal, tangent), (jnp_primal, jnp_tangent) = _make_primal_tangents(mode)
         (out, t_out), (minus_out, tt_out) = linear_solve3(primal, tangent)
@@ -708,7 +712,9 @@ def jvp_jvp_impl(
             jnp_primal, jnp_tangent
         )
 
-        assert tree_allclose(out, true_out, atol=1e-4)
-        assert tree_allclose(t_out, true_t_out, atol=1e-4)
-        assert tree_allclose(tt_out, true_tt_out, atol=1e-4)
-        assert tree_allclose(minus_out, minus_true_out, atol=1e-4)
+        assert tree_allclose(eqx.combine(out, true_out), true_out, atol=1e-4)
+        assert tree_allclose(eqx.combine(t_out, true_t_out), true_t_out, atol=1e-4)
+        assert tree_allclose(eqx.combine(tt_out, true_tt_out), true_tt_out, atol=1e-4)
+        assert tree_allclose(
+            eqx.combine(minus_out, minus_true_out), minus_true_out, atol=1e-4
+        )
--- a/tests/helpers.py
+++ b/tests/helpers.py
@@ -29,71 +29,135 @@ from equinox.internal import ω
 
 @ft.cache
 def _construct_matrix_impl(
-    getkey, tags, size, dtype, cond_or_singular: int | float | str, i: int
+    getkey,
+    tags,
+    rows: int,
+    cols: int,
+    full_rank: bool,
+    dtype,
+    cond_cutoff: float | None,
+    i: int,
 ):
-    del i  # used to break the cache
-    while True:
-        matrix = jr.normal(getkey(), (size, size), dtype=dtype)
-        if isinstance(cond_or_singular, str):
-            if cond_or_singular == "zero":
-                matrix = matrix.at[0, :].set(0)
-            elif cond_or_singular == "trim_row":
-                matrix = matrix[1:, :]
-            elif cond_or_singular == "trim_col":
-                matrix = matrix[:, 1:]
-        if tags != ():
-            assert (
-                isinstance(cond_or_singular, (int, float)) or cond_or_singular == "zero"
+    if i > 0:  # we are giving a tangent vector
+        # handle the tags special cased for fast mv by _try_sparse_materialise,
+        # as tangent vectors must also have the tags to be mv'd correctly.
+
+        # It is otherwise not necessary that tangent vectors (and higher order)
+        # have the same tags as the primal unless we have optimized mv to
+        # exploit the structure of the tags. Furthermore, we pick tangent
+        # directions tangent to the constant rank locus at the primal.
+        t_matrix = jr.normal(getkey(), (rows, cols), dtype=dtype)
+        if not full_rank:
+            matrix = _construct_matrix_impl(
+                getkey, tags, rows, cols, full_rank, dtype, cond_cutoff, 0
             )
+            r = min(rows, cols) - 1
+            t_matrix = jax.jvp(lambda A: Ahat(A, r), (matrix,), (t_matrix,))[1]
         if has_tag(tags, lx.diagonal_tag):
-            matrix = jnp.diag(jnp.diag(matrix))
-        if has_tag(tags, lx.symmetric_tag):
-            matrix = matrix + matrix.T
+            t_matrix = jnp.tril(jnp.triu(t_matrix))
+            assert full_rank
+        if has_tag(tags, lx.tridiagonal_tag):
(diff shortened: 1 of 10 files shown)


tests/test_well_posed.py
```diff
@@ -52,35 +52,46 @@ def test_small_wellposed(make_operator, solver, tags, ops, getkey, dtype):
...
+        lx.LU(),
+        lx.QR(),
+        lx.SVD(),
+        lx.BiCGStab(rtol=tol, atol=tol),
+        lx.GMRES(rtol=tol, atol=tol),
+        lx.Normal(lx.CG(rtol=tol, atol=tol)),
+        lx.LSMR(atol=tol, rtol=tol),
+        lx.Normal(lx.Cholesky()),
+    ),
+)
```
jpbrodrick89, line 70: why are we hardcoding here rather than e.g. using solver_features directly, this is a future footgun meaning that we might forget to add new solvers.

tests/test_singular.py
```diff
@@ -228,43 +215,44 @@ def test_nonsquare_vec(solver, full_rank, jvp, wide, dtype, getkey):
...
-
-    if use_state:
-        state = solver.init(operator, options={})
-        linear_solve = ft.partial(lx.linear_solve, state=state)
-    else:
-        linear_solve = lx.linear_solve
-
-    with pytest.raises(Exception):
-        linear_solve(operator, vec, solver)
+# _iterative_solvers = (
```
jpbrodrick89, line 232: why is this commented out, if its covered elsewhere or a bad test (i.e. we can't guarantee an exception will be raised) just remove the test and have a single comment explaining why this test was remove as part of this PR. (e.g. `We don't test ... because .... Test removed in PR ..."

tests/test_well_posed.py
```diff
@@ -52,35 +52,46 @@ def test_small_wellposed(make_operator, solver, tags, ops, getkey, dtype):
...
 @pytest.mark.parametrize("dtype", (jnp.float64, jnp.complex128))
 def test_pytree_wellposed(solver, getkey, dtype):
-    if not isinstance(
-        solver,
-        (lx.Diagonal, lx.Triangular, lx.Tridiagonal, lx.Cholesky, lx.CG),
-    ):
```
jpbrodrick89, line 61: Why do we want to add these back in? Tridagonal specifically makes very little sense for a pytree operator.  I would probably argue the Cholesky/CG doesn't really either (what does it mean for a pytree operator to be spd?). I'd be most tempted to keep diagonal and triangular (these make sense, especially if the matrix is materialised in a deterministic order) but I'm not sure about the rest?

tests/test_vmap_jvp.py
```diff
@@ -22,112 +22,147 @@
...
+            linear_solve2 = lambda vector: linear_solve1(operator, vector)
+            jnp_solve2 = lambda vector: jnp_solve1(matrix, vector)
+        elif mode == "op_vec":
+            linear_solve2 = linear_solve1
+            jnp_solve2 = jnp_solve1
+        else:
+            assert False
+        jnp_solve3 = ft.partial(eqx.filter_jvp, jnp_solve2)
+        jnp_solve3 = eqx.filter_vmap(jnp_solve3)
+        jnp_solve3 = eqx.filter_jit(jnp_solve3)
```
jpbrodrick89, line 146: Is there a particular reason for this refactor? Might jnp_solve3 be used if jvp_first?

tests/test_vmap_jvp.py
```diff
@@ -22,112 +22,147 @@
...
+            return lx.linear_solve(operator, vector, state=state, solver=solver)
+
+    else:
+        linear_solve1 = ft.partial(lx.linear_solve, solver=solver)
+
+    if full_rank and rows == cols:
+        jnp_solve1 = jnp.linalg.solve
+    else:
+        r = min(rows, cols) if full_rank else min(rows, cols) - 1
+        jnp_solve1 = lambda mat, vec: jnp_lstsq(mat, vec, r)
```
jpbrodrick89, line 90: nit: In general, I'd prefer to use partial (ambivalent whether to use functools/jax/equinox) for NEW lambda functions (if you are refactoring a previous lambda that's fine)

tests/test_jvp.py
```diff
@@ -21,100 +21,95 @@
...
-        operator, t_operator = eqx.filter_jvp(
-            make_op, (matrix, tags), (t_matrix, t_tags)
-        )
+    if has_tag(tags, lx.unit_diagonal_tag):
+        # For all the other tags, A + εB with A, B \in {matrices satisfying the tag}
+        # still satisfies the tag itself.
+        # This is the exception.
+        t_matrix = t_matrix.at[
+            jnp.arange(min(rows, cols)), jnp.arange(min(rows, cols))
+        ].set(0)
```
jpbrodrick89, line 72: is there any reason why we wouldn't do this in construct_matrix_impl on the i > 0 branch where we already doing zeroing work for the diagonal and off diagonal case?

tests/test_vmap_jvp.py
jpbrodrick89: sorry, I was reading the "ignore whitespace" diff which made it look like jnp_solve3 was only called `if not jvp_first` but apparently its exacuted on both branches, I see that this should be a slight performance improvement.

tests/helpers.py
```diff
@@ -29,71 +29,135 @@
...
+        "Cannot construct rank deficient matrix with bounded condition number"
+    )
+    while True:
+        matrix = jr.normal(getkey(), (rows, cols), dtype=dtype)
+        if isinstance(tags, tuple) and len(tags) == 0 and not full_rank:
+            u, s, vh = jnp.linalg.svd(matrix, full_matrices=False)
+            s = s.at[-1].set(0).astype(matrix.dtype)
+            matrix = (u * s[None, :]) @ vh
+        if has_tag(tags, lx.diagonal_tag):
+            matrix = jnp.tril(jnp.triu(matrix))
```
jpbrodrick89, line 73: is the tril+triu approach more performant than the original diag+diag approach?

tests/helpers.py
```diff
@@ -29,71 +29,135 @@
...
-            assert (
-                isinstance(cond_or_singular, (int, float)) or cond_or_singular == "zero"
+    if i > 0:  # we are giving a tangent vector
+        # handle the tags special cased for fast mv by _try_sparse_materialise,
+        # as tangent vectors must also have the tags to be mv'd correctly.
+
+        # It is otherwise not necessary that tangent vectors (and higher order)
+        # have the same tags as the primal unless we have optimized mv to
+        # exploit the structure of the tags. Furthermore, we pick tangent
+        # directions tangent to the constant rank locus at the primal.
```
jpbrodrick89, line 48: I don't really understand this. Doesn't lineax implicitly assume that a tangent operator preserve the tags of its primal? For example, low rank + projectd low rank = low rank, diag + diag = diag, tridiag + tridiag, unit diag + zero diag = unit diag (we do all of these). Why don't we need to do symm + symm = symm, spd + spd = spd, tri + tri = tri, etc.?

tests/helpers.py
```diff
@@ -29,71 +29,135 @@
...
+            matrix = matrix + matrix.T
+            if not full_rank:
+                matrix = matrix.at[0, :].set(0).at[:, 0].set(0)
         if has_tag(tags, lx.positive_semidefinite_tag):
-            matrix = matrix @ matrix.T.conj()
+            assert rows == cols, "Positive semidefinite matrix must be square"
+            u, s, _ = jnp.linalg.svd(matrix)
+            if not full_rank:
+                s = s.at[-1].set(0)
+            matrix = (u * s[None, :].astype(matrix.dtype)) @ u.conj().T
```
jpbrodrick89, line 106: this is likely an important change but non-intuitive. Explain in a comment why its ok to do `matrix.at[0, :].set(0).at[:, 0].set(0)` for symmetric and not for spd (I'm guessing it does guarantee spd'ness).

tests/helpers.py
```diff
@@ -334,6 +421,117 @@ def make_composed_operator(getkey, matrix, tags):
...
+            return jnp.linalg.solve(A.conj().T @ A, A.conj().T @ v)
+        else:
+            return A.conj().T @ jnp.linalg.solve(A @ A.conj().T, v)
+    else:
+        # We compute lstsq(B @ C, v) as lstsq(C, lstsq(B, v)), but use solve with
+        # the normal equations to avoid implicitly differentiating through an svd
+        # solve
+        B, C = Ahat_factor(A, r)
+        y = jnp.linalg.solve(B.conj().T @ B, B.conj().T @ v)
+        x = C.conj().T @ jnp.linalg.solve(C @ C.conj().T, y)
```
jpbrodrick89, line 531: consider recursively calling jnp_lstsq as probably more transparent and does exactly the same thing

tests/helpers.py
```diff
@@ -334,6 +421,117 @@ def make_composed_operator(getkey, matrix, tags):
...
+        C = vh[:r, :]
+        Cfree = jnp.linalg.solve(C[:, :r], C[:, r:])
+        return Cfree
+
+    @Cfree.defjvp
+    def Cfree_jvp(primals, tangents):
+        # obtain derivatives by the implicit function theorem applied to the
+        # conditions defining Cfree. Everything here must be differentiable
+        # in order for higher derivatives to be defined: Cfree is obtained
+        # recursively and everything else is obtained with autodifferentiable
```
jpbrodrick89, line 447: is "recursive" the right term here? Cfree is only called once (on line 451), right? That's just a "forward pass". I would rename all the Cfree args below as easy to get mixed up/confused.

tests/helpers.py
```diff
@@ -334,6 +421,117 @@ def make_composed_operator(getkey, matrix, tags):
...
 
 
+# Nearest rank r factorization via Eckart-Young-Mirsky + IFT.
+# Computes B, C so that Ahat = B @ C where the inner dimension is r,
+# C[:,:r] = I_r, and |A - Ahat|_F is smallest.
+# The JVP is defined via the implicit function theorem applied to the
+# stationarity conditions of the Frobenius-norm minimization on the free
+# variables of C (B obtained by linear least squares).
+# This is smooth in a neighborhood of any A where s_r is separated from
+# s_{r+1} and A[:,:r] is full rank
```
jpbrodrick89, line 431: I think A[:,:r] -> vh[:r, :r]

jpbrodrick89 (commented):
Sorry this took so long to get to @adconner, ignoring whitespace made the changes a bit easier to digest! 😅 I like the essence of the refactor, but am a bit concerned about the complexity of Ahat_factor and maintaining it going forward. Are there simpler alternatives worth considering.

The tests don't need the _closest_ rank-r approximation. They need:

* A rank-r matrix (primal)
* A tangent vector that lies in the tangent space of the rank-r manifold at that matrix

The Eckart-Young-Mirsky theorem is used to project random matrices onto the rank-r manifold. But we could just generate directly on the manifold from the start.

i.e.

```python
r = min(rows, cols) - 1

# Primal
B = jr.normal(getkey(), (rows, r), dtype=dtype)
C = jr.normal(getkey(), (r, cols), dtype=dtype)
matrix = B @ C

# Tangent — guaranteed in tangent space, no SVD needed
dB = jr.normal(getkey(), (rows, r), dtype=dtype)
dC = jr.normal(getkey(), (r, cols), dtype=dtype)
t_matrix = dB @ C + B @ dC
```

This should mean we can also differentiate through the double normal solve directly without a custom jvp.

I'm also slightly hesitant about dropping finite differences when what we are testing is slightly opaque (i.e. custom jvp's), finite differences is a nice simple (but of course low accuracy) comparison that seems perhaps dangerous to drop.

Note this seems to add almost 40 minutes onto test runtime, do you think the increased surface area is worth it (particularly vmap, and vmap of vmap are unlikely to benefit from such a wide surface area)? If you think its critical, we could maybe consider adding a slow tag to some "overkill" tests so that new devs know what the core tests are. I find that the longer tests take, the less likely it is that devs will bother to run the whole suite meaning that problems may only be covered once the PR is made and tests are run there.

Finally the jnp_lstsq helper function is misleadingly named and confused me to start with, something like solve_maybe_normal_rank_preserving, lstsq_with_jnp_solve or something similar might be more intuitive. There should also be some more comments in the function on why you are solving normal equations and not using the lstsq even in the (purportedly) full rank case.

Happy to discuss further especially if I've missed something or solicit @patrick-kidger's opinion as he is much more familiar with this part of the test suite than me.

Thanks again for the contribution!

tests/test_well_posed.py
adconner: My thinking was  maybe it pays to be a bit more explicit, since we need to drop tags (and deduplicate solvers) as well as remove (lx.Diagonal, lx.Triangular, lx.Tridiagonal, lx.Cholesky, lx.CG), so maybe just enumerate the examples wellposed pytree applies to. I'll rewrite this logic using the solver features

tests/test_singular.py
adconner: I commented it because I wanted to have a discussion. I think it should be deleted as we cannot guarantee this error is raised. This is specifically needed with this change as we are now testing against random singular matrices to machine precision rather than structural (zero row or column, eg), so with much higher probability these solvers actually succeed (and thus fail the test). Should the discussion you suggest be made in the commit message or the code?

tests/test_well_posed.py
adconner: I agree. The tested solvers were changed to an explicit list omitting these examples.

tests/test_jvp.py
adconner: This was left in as a mistake, I'll fix. Its actually unnecessary for the tangents, including this one, to satisfy the primal tags.

tests/helpers.py
adconner: I did not measure this. My expectation was slightly so, as the computational graph just becomes pass through the entries we want, zeroing the others, rather than do three diagonal extractions, diagonal matrix formation and sum. I don't have a strong opinion on which method is taken.

tests/helpers.py
adconner: I dont believe so. One can view the tags as guarantees which mainly allow solver optimizations, and the jvp rule never needs to solve with respect to a tangent vector. So in principle a user can, supposing they want to differentiate through a solve A(t)x = b at t=0, change basis so that A(0) is very simple, maybe triangular or diagonal, only differentiate through this very simple kind of solve, and then change base back to the desired basis. Of course this requires that one can use an arbitrary tangent vector, something which our jvp rule actually supports. 

The only slightly sketchy thing is that internally the operators representing the tangents have the same tags as the primal, even if could logically represent in some contexts arbitrary matrices. Currently tags only matter when you want to solve or if you want to mv and the tags include diagonal or tridiagonal, where now mv exploits the structure for speed.

As a result, this potential functionally is not documented and maybe not officially supported. But my idea is we keep our tests general until we tangent tags are actually used for optimizations, probably for mv.

tests/helpers.py
adconner: In my mind its more that its hard to actually generate a random rank deficient symmetric matrix which is not necessarily definite, so I resort to structural zeroing. Actually thinking about it now, I could structurally zero and then make a random change of variables to get a random symmetric matrix, not necessarily definite. Random rank deficient symmetric matrix is currently unused anywhere, so I did not think too hard how to generate them. 

In the definite case, I can zero an eigenvalue of a random definite matrix to get a random rank deficient.

tests/helpers.py
adconner: Fair enough. In my mind I was trying to figure out how to make this function infinitely differentiable, and the solution is that the jvp of the jvp becomes recursive by calling the primal again whose derivative is currently being defined. But I can remove this language.

tests/helpers.py
adconner: These are equivalent statements. I express using A so that the conditions on the arguments are clear.

adconner:
> The tests don't need the _closest_ rank-r approximation. They need:
> 
>     * A rank-r matrix (primal)
> 
>     * A tangent vector that lies in the tangent space of the rank-r manifold at that matrix
> 
> 
> The Eckart-Young-Mirsky theorem is used to project random matrices onto the rank-r manifold. But we could just generate directly on the manifold from the start.
> 
> i.e.
> 
> ```python
> r = min(rows, cols) - 1
> 
> # Primal
> B = jr.normal(getkey(), (rows, r), dtype=dtype)
> C = jr.normal(getkey(), (r, cols), dtype=dtype)
> matrix = B @ C
> 
> # Tangent — guaranteed in tangent space, no SVD needed
> dB = jr.normal(getkey(), (rows, r), dtype=dtype)
> dC = jr.normal(getkey(), (r, cols), dtype=dtype)
> t_matrix = dB @ C + B @ dC
> ```

I began by trying to develop this approach. While it probably could be made to work, it requires much more invasive changes to the tests, and much of the test complexity then becomes unrelated to the tested code (all tested matrices and tangents are given by explicit factorizations, etc). Additionally, I think I ran into issues correctly testing the second derivatives correctly. Once I realized it was possible to just fix jax's lstsq I did that. 

> 
> This should mean we can also differentiate through the double normal solve directly without a custom jvp.
> 
> I'm also slightly hesitant about dropping finite differences when what we are testing is slightly opaque (i.e. custom jvp's), finite differences is a nice simple (but of course low accuracy) comparison that seems perhaps dangerous to drop.

This is reasonable. One still needs to be careful implementing finite differences with rank deficient solves though, as A^+ is discontinuous near the rank deficient locus. (This didnt show up before because we were testing only at very special matrices which were rank deficient due to structural zeros, eg, and we chose the tangents to have the same structural zeros thus never leaving the rank r locus by construction when doing finite differences. But doing this lead to significant loss in generality in the tested situations) To do it correctly you still need to logically do finite differences on the function I call jnp_lstsq, as this is the function our jvp rule implements. But sure we could replace the analytic jvp with a finite differences jvp. In my mind once I realized it was possible to define the reference jvp analytically this was worthwhile but maybe this is a matter of taste. 

> 
> Note this seems to add almost 40 minutes onto test runtime, do you think the increased surface area is worth it (particularly vmap, and vmap of vmap are unlikely to benefit from such a wide surface area)? If you think its critical, we could maybe consider adding a slow tag to some "overkill" tests so that new devs know what the core tests are. I find that the longer tests take, the less likely it is that devs will bother to run the whole suite meaning that problems may only be covered once the PR is made and tests are run there.

Yes this is a downside. Let me investigate again where most of the time is being taken to see if what is added is really necessary.

> 
> Finally the jnp_lstsq helper function is misleadingly named and confused me to start with, something like solve_maybe_normal_rank_preserving, lstsq_with_jnp_solve or something similar might be more intuitive. There should also be some more comments in the function on why you are solving normal equations and not using the lstsq even in the (purportedly) full rank case.

jnp_lstsq is just supposed to be jnp.lstsq but where derivatives are correct and not NaN. The only interface difference is the r parameter, but honestly I can remove that and make the interface identical. 

I solve the normal equations even when full rank because we have to work around situations where any singular values are repeated, not just zero singular values. This shows up when trying to differentiate solves through the identity matrix, eg. But as you say I can probably simplify by recursively calling jnp_lstsq.

tests/helpers.py
adconner: I guess technically its Ahat[:,:r] full rank which is really equivalent, so point taken

jpbrodrick89:
> I began by trying to develop this approach. While it probably could be made to work, it requires much more invasive changes to the tests, and much of the test complexity then becomes unrelated to the tested code (all tested matrices and tangents are given by explicit factorizations, etc). Additionally, I think I ran into issues correctly testing the second derivatives correctly. Once I realized it was possible to just fix jax's lstsq I did that.

Thanks, if that's the case happy to go with this. I've been working on a projection operator and your new tests caught second order derivative bug (also present in draft #217 which I will likely close, the projection part is handled by the new work). So I'm very keen on getting it in now I can see the real value the extra tests add.

I don't know if this helps at all, but Claude gave the following suggestion for getting the direct manifold generation approach to work.

```python
@ft.cache
def _rank_deficient_factors(getkey, rows, cols, r, dtype, seed):
    """All factors needed for num=4, generated once and cached."""
    B  = jr.normal(getkey(), (rows, r), dtype=dtype)
    C  = jr.normal(getkey(), (r, cols), dtype=dtype)
    dB = jr.normal(getkey(), (rows, r), dtype=dtype)
    dC = jr.normal(getkey(), (r, cols), dtype=dtype)
    eB = jr.normal(getkey(), (rows, r), dtype=dtype)
    eC = jr.normal(getkey(), (r, cols), dtype=dtype)
    return B, C, dB, dC, eB, eC

def _get_rank_deficient_matrix(i, B, C, dB, dC, eB, eC):
    if i == 0: return B @ C
    if i == 1: return dB @ C + B @ dC
    if i == 2: return eB @ C + B @ eC
    if i == 3: return eB @ dC + dB @ eC   # cross-term
```

It is very possible though that the second order derivative just doesn't comply with this approach as it might be inconsistent with the EYM approach. Also, it might be worth considering does the fact that pivoted QR has a different retraction mapping mean that it might fail these tests, worth sense checking before we buy into it.

I think you're right about the tangent tags stuff and was a subtle thing I completely missed, I think not guaranteeing to handle tangents without the same tags is safest for now (as in the fast mv path cases) but testing it to show regression in the currently supported cases when it breaks is still helpful.

I've merged #225 in dev/  so if you can rebase off there that would be much appreciated. If you want to parameterise over max rank tag in a useful way note this will only be supported by `assume_full_rank=False` solvers but `test_max_rank` might already handle all we need.

Let me know if you need any assistance, thanks so much for your help wit this!

PR #222, opened 2026-05-13
stop Mul/DivLinearOperator converting static scalars to jax arrays

jpbrodrick89:
I realised that our psd/nsd tags weren't propagating with simple primitive multipliers, the culprit was simply `jnp.asarray` converting scalar to jax arrays which escaped detection, using `np.ndim` works in all cases.

The one things I'm not really sure about is whether it was worth adding the unit diagonal `1 *` and `/ 1` tests, it seems a bit niche but probably harmless. Happy to remove or keep as you prefer.

The way I updated the tests to catch this is somewhat nuanced and I'd appreciate a look whenever you get a small moment (no rush on this one!).

### DIFF
--- a/lineax/_operator.py
+++ b/lineax/_operator.py
@@ -34,7 +34,7 @@ from jaxtyping import (
     ArrayLike,
     Inexact,
     PyTree,  # pyright: ignore
-    Scalar,
+    ScalarLike,
     Shaped,
 )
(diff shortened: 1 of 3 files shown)


patrick-kidger:
If I'm reading this correctly then I think your changes LGTM!

PR #224, opened 2026-05-26
refactor: create invert tags helper

jpbrodrick89:
This mirrors the existing transpose_tags helper and simplifies the logic in _solve.py. The main goal is unifying tags logic so if a developer adds a new tag (e.g. the proposed `circulant_tag`  from discussions with @BalzaniEdoardo) they don't need to trace as much of the codebase. The tags_from_checks helper is the most janky due to its circular import with `_operatory.py` (solved with lazy importing and `TYPE_CHECKING` conditionals). Happy to move that to `_operator.py` if you prefer.

### DIFF
--- a/lineax/__init__.py
+++ b/lineax/__init__.py
@@ -69,10 +69,13 @@ from ._solver import (
 )
 from ._tags import (
     diagonal_tag as diagonal_tag,
+    invert_tags as invert_tags,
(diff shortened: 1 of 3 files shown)


patrick-kidger:
This LGTM I think! No strong feelings on jankiness :)

PR #228, opened 2026-06-02
refactor _operator.py into multiple files

jpbrodrick89:
Implements #227 breaking the ~2400-line _operator.py into a _operator/ package:

- **base.py**: `AbstractLinearOperator` and the `singledispatch` operation stubs (558 lines)
- **core.py**: Matrix, PyTree, Jacobian, Function (819 lines)
- **structured.py:** Identity, Diagonal, Tridiagonal (382 lines)
- **wrapper.py**: Tagged, Tangent, Mul, Neg, Div (524 lines)
- **binary.py**: Add, Composed (277 lines)
- `__init__.py` re-exports the public API unchanged (53 lines)

Operation registrations are co-located with their operators, and repeated registrations are folded into loops (the core predicates and the structured checks). Cross-file helpers drop their leading underscore (with `_frozenset` renamed to `as_frozenset`), and `_try_sparse_materialise` becomes `try_structured_materialise`. structured.py is ordered Identity -> Diagonal -> Tridiagonal, and the **operator API docs** are regrouped into core / structured / wrapper to match. (@patrick-kidger the docs restructure is probably the thing you might be most opinionated about so I will hold off merging until you've confirmed comfort with that.)

Behaviour-preserving: the public imports are unchanged, and the operator, transpose, and adjoint test suites pass (full suite will run in CI).

@BalzaniEdoardo a new `CirculantLinearOperator.py` can now fit cleanly into `structured.py` which is only 382 lines after this refactor.

PR #230, opened 2026-06-05
fix iterative breakdowns due to element-wise tolerances

jpbrodrick89:
A flaky GMRES test on #228 led me down a rabbit hole to realise that all iterative solvers except for LSMR have been using nonlinear solver style element-wise convergence checks (e.g. Hairer–Nørsett–Wanner) to handle elements with hugely differing scale, i.e.:

```python
b_scale = (self.atol + self.rtol * ω(vector).call(jnp.abs)).ω
y_scale = (self.atol + self.rtol * ω(y).call(jnp.abs)).ω
norm1 = self.norm((r**ω / b_scale**ω).ω)
norm2 = self.norm((diff**ω / y_scale**ω).ω)
return (norm1 > 1) | (norm2 > 1)
```

However, for iterative LINEAR solvers RHS elements of different scales will inevitably mix unless the matrix is sufficiently sparse leading to round-off error that to "iterative breakdowns". These cases can crop up randomly in the tests.

MWE:

```python
A = jnp.array([[1.0, 0.5, 0.3],
               [0.2, 2.0, 0.7],
               [0.6, 0.1, 1.5]])
b = jnp.array([1e8, 1.0, 1.0])

lx.linear_solve(lx.MatrixLinearOperator(A), b, lx.GMRES(rtol=1e-12, atol=1e-12))
# -> equinox.EquinoxRuntimeError: A form of iterative breakdown has occurred ...

# ...even though the system is solved essentially perfectly:

sol = lx.linear_solve(lx.MatrixLinearOperator(A), b,
                      lx.GMRES(rtol=1e-12, atol=1e-12), throw=False)
print(jnp.linalg.norm(A @ sol.value - b) / jnp.linalg.norm(b))   # 1.7e-16
print(jnp.allclose(sol.value, jnp.linalg.solve(A, b)))           # True
```

Note that the same solve passes at rtol=atol=1e-6 — the bug only bites when the requested tolerance dips near the achievable round-off floor.

This is fixed by adopting the standard convention (e.g. as in scipy) to only compare the norm of the vector to the tolerance:

```python
b_scale = self.atol + self.rtol * self.norm(vector)
y_scale = self.atol + self.rtol * self.norm(y)
return (self.norm(r) > b_scale) | (self.norm(diff) > y_scale)
```

@patrick-kidger this has been hugely long-standing (introduced in "initial commit") and I understand you might have some affinity towards the original approach. As such, I will await your oversight before merging this.

### DIFF
--- a/lineax/_solver/bicgstab.py
+++ b/lineax/_solver/bicgstab.py
@@ -93,7 +93,7 @@ class BiCGStab(AbstractLinearSolver[_BiCGStabState]):
             and self.rtol == 0
         )
         if has_scale:
-            b_scale = (self.atol + self.rtol * ω(vector).call(jnp.abs)).ω
+            b_scale = self.atol + self.rtol * self.norm(vector)
 
         # This implementation is the same a jax.scipy.sparse.linalg.bicgstab
         # but with AbstractLinearOperator.
@@ -117,11 +117,16 @@ class BiCGStab(AbstractLinearSolver[_BiCGStabState]):
             # Given Ay=b, then we have to be doing better than `scale` in both
             # the `y` and the `b` spaces.
             if has_scale:
-                with jax.numpy_dtype_promotion("standard"):
-                    y_scale = (self.atol + self.rtol * ω(y).call(jnp.abs)).ω
-                    norm1 = self.norm((r**ω / b_scale**ω).ω)  # pyright: ignore
-                    norm2 = self.norm((diff**ω / y_scale**ω).ω)
-                return (norm1 > 1) | (norm2 > 1)
+                # Standard relative-residual stopping rule: ‖r‖ ≤ atol + rtol·‖b‖
+                # (and likewise for the increment in the `y` space). Note this uses
+                # scalar norms, *not* an elementwise `atol + rtol·|b|` scale: the
+                # latter is unsatisfiable for wide-dynamic-range `b`, where the
+                # round-off floor of large components exceeds the absolute tolerance
+                # demanded of small ones, yielding spurious non-convergence.
+                y_scale = self.atol + self.rtol * self.norm(y)
+                b_unconverged = self.norm(r) > b_scale  # pyright: ignore
+                y_unconverged = self.norm(diff) > y_scale
+                return b_unconverged | y_unconverged
             else:
                 return True
 
--- a/lineax/_solver/cg.py
+++ b/lineax/_solver/cg.py
@@ -17,7 +17,6 @@ from collections.abc import Callable
 from typing import Any, TypeAlias
 
 import equinox.internal as eqxi
-import jax
 import jax.lax as lax
 import jax.numpy as jnp
 import jax.tree_util as jtu
@@ -144,18 +143,23 @@ class CG(AbstractLinearSolver[_CGState]):
             and self.rtol == 0
         )
         if has_scale:
-            b_scale = (self.atol + self.rtol * ω(vector).call(jnp.abs)).ω
+            b_scale = self.atol + self.rtol * self.norm(vector)
 
         def not_converged(r, diff, y):
             # The primary tolerance check.
             # Given Ay=b, then we have to be doing better than `scale` in both
             # the `y` and the `b` spaces.
             if has_scale:
-                with jax.numpy_dtype_promotion("standard"):
-                    y_scale = (self.atol + self.rtol * ω(y).call(jnp.abs)).ω
-                    norm1 = self.norm((r**ω / b_scale**ω).ω)  # pyright: ignore
-                    norm2 = self.norm((diff**ω / y_scale**ω).ω)
-                return (norm1 > 1) | (norm2 > 1)
+                # Standard relative-residual stopping rule: ‖r‖ ≤ atol + rtol·‖b‖
+                # (and likewise for the increment in the `y` space). Note this uses
+                # scalar norms, *not* an elementwise `atol + rtol·|b|` scale: the
+                # latter is unsatisfiable for wide-dynamic-range `b`, where the
+                # round-off floor of large components exceeds the absolute tolerance
+                # demanded of small ones, yielding spurious non-convergence.
+                y_scale = self.atol + self.rtol * self.norm(y)
+                b_unconverged = self.norm(r) > b_scale  # pyright: ignore
+                y_unconverged = self.norm(diff) > y_scale
+                return b_unconverged | y_unconverged
             else:
                 return True
 
--- a/lineax/_solver/gmres.py
+++ b/lineax/_solver/gmres.py
@@ -116,7 +116,7 @@ class GMRES(AbstractLinearSolver[_GMRESState]):
             and self.rtol == 0
         )
         if has_scale:
-            b_scale = (self.atol + self.rtol * ω(vector).call(jnp.abs)).ω
+            b_scale = self.atol + self.rtol * self.norm(vector)
         operator = state
         preconditioner, y0 = preconditioner_and_y0(operator, vector, options)
         leaves, _ = jtu.tree_flatten(vector)
@@ -132,11 +132,16 @@ class GMRES(AbstractLinearSolver[_GMRESState]):
             # Given Ay=b, then we have to be doing better than `scale` in both
             # the `y` and the `b` spaces.
             if has_scale:
-                with jax.numpy_dtype_promotion("standard"):
-                    y_scale = (self.atol + self.rtol * ω(y).call(jnp.abs)).ω
-                    norm1 = self.norm((r**ω / b_scale**ω).ω)  # pyright: ignore
-                    norm2 = self.norm((diff**ω / y_scale**ω).ω)
-                return (norm1 > 1) | (norm2 > 1)
+                # Standard relative-residual stopping rule: ‖r‖ ≤ atol + rtol·‖b‖
+                # (and likewise for the increment in the `y` space). Note this uses
+                # scalar norms, *not* an elementwise `atol + rtol·|b|` scale: the
+                # latter is unsatisfiable for wide-dynamic-range `b`, where the
+                # round-off floor of large components exceeds the absolute tolerance
+                # demanded of small ones, yielding spurious non-convergence.
+                y_scale = self.atol + self.rtol * self.norm(y)
+                b_unconverged = self.norm(r) > b_scale  # pyright: ignore
+                y_unconverged = self.norm(diff) > y_scale
+                return b_unconverged | y_unconverged
             else:
                 return True
 


patrick-kidger:
So this is actually an explicit choice, and we have some discussion on this in Diffrax [here](https://github.com/patrick-kidger/diffrax/blob/2cd98903d051d2f11ec47fda15f2d190095392d6/diffrax/_step_size_controller/pid.py#L412). From the discussion there, I'd actually have thought that the current formulation should be more stable when considering components with different scales?

Basically, if we ever do `norm(a) / norm(b)` then we'll end up with a weaker convergence condition – we just wash away all the small scales and only really care if the big ones have converged. That doesn't seem particularly fair to the small scales.

It sounds like your MWE represents a case where the small scales are now being *overrepresented* instead? Perhaps due to numerical stability/precision issues?

(Side note, one of the things we did try to do across Lineax/Optimistix/Diffrax was standardise convergence criteria on the one we have here, as broadly speaking the rest of the numerical literature just does in some ad-hoc problem-specific way every time.)

jpbrodrick89:
> Perhaps due to numerical stability/precision issues?

Yes, it is entirely because of numerical precision/round-off error, I believe GMRES is guaranteed to converge in n iterations with a tolerance of 0 if we have infinite precision.

The point is that is A.mv couples components of different scales. This sets a tolerance floor of order eps * cond(A) * max(b) (this is very approximate, the true floor is probably better expressed the way LSMR handles it).

I would argue that iterative linear solvers only ever make theoretical guarantees about convergence in the Euclidean norm (this is the "minimum residual" referred to in GMRES) and element-wise covergence is an unrealistic expectation. On the other hand failure of an ODE nonlinear solve allows a resolution through adaptive stepping (it is essentially a signal that a problem is too "difficult" to solve in some nonlinear sense and dropping the timestep will make it more linear, if a problem is already linear then no scaling would make an iterative solver behave any better).

That said if this is too much of a paradigm shift we can just fix the flaky tests by patching this into atol (i.e. atol=eps*max_cond*max(b)~1e-13*max(b) which turns to be 1e-8/1e-9 when using calculating a jvp (as max(t_matrix@b) scales as cond(t) max(b)) the current tol is probably fine for the primal solve. So in summary, reducing atol by ~1e3-4 for test_jvp and another factor again for test_jvp_jvp wil probably cut it but in my opinion its a bit awkward.

patrick-kidger:
> I would argue that iterative linear solvers only ever make theoretical guarantees about convergence in the Euclidean norm (this is the "minimum residual" referred to in GMRES)

I think that's a reasonable position to take. In theory convergence in one norm guarantess convergence in all others, but as a practical matter that's not really how things work :D

I also like your observation that generally speaking failures become a matter of "can we make this problem more linear".

---

One other guiding principle that went into this, btw, is to ask what would happen if we pad our problems with extra zeros. Ideally the result would be identical. (And this principle is also the reason that Lineax unifies the finding of inverses and pseudoinverses, e.g. consider adding extra zeros to both operator and vector in a well-posed problem: the resulting pseudoinverse solution is just the similarly-padded inverse solution of the original problem.)

It's for this reason that we use `norm=max_norm` pretty much ubiquitously, and not `two_norm`.  
(Although checking this over I've just spotted that `LSMR` uses `two_norm`, that snuck by me... so whatever we decide here we should make sure it's consistent over there too.)

Do you think there is any way we could keep your "make this problem more linear' intuition, whilst preserving this kind of zero-padding property?

jpbrodrick89:
I'm not sure I follow, I think the result remains identical under zero padding. I'm not proposing to change norm but I think both max_norm and two_norm have this property (this is because both norms themselves are invariant under zero padding). The thing that this PR DOES affect is behaviour under block diagonal padding. After this PR we would likely terminate sooner before one of the block diagonals has converged because we don't know a priori there is no mixing that would imply we COULDN'T converge numerically.

jpbrodrick89:
Following up from our conversation on Monday, I don't there exists a general norm that would give us the same results as before that we could offer as an alternative. If we want to allow users to recover the original we could offer an `elementwise_rtol=None` option (we assume they would want to use the same `atol` elementwise and globally even though that would mean `atol` is more lax in the elementwise setting than the norm-based one). To reiterate, I think its safe to assume that repeated mv's will essentially mix all elements of the vector rather than try come up with janky tag-based exceptions to this until we eventually identify an elegant sparse/block operator design.

jpbrodrick89:
I think I've worked out the "elementwise" norm we desired that would recover the current behaviour is just a _weighted_ norm with weights that depend on y/vector. So one solution is to allow/enforce a second argument on our norm functions, e.g.

```python
def elementwise_norm(x, args):
    atol, rtol, ref = args
    return max_norm((x**ω / (atol + rtol * ω(ref).call(jnp.abs)).ω)
```

We would need to refactor all our existing norms to take these dummy args.

Alternatively one could hardcode some/all of the args but that's probably less flexible.

WDYT? Default to standard max_norm as this PR suggests and add the dummy args and elementwise_norm functionality to allow recovery of previous behaviour?

patrick-kidger:
Ah, interesting – you're saying that you've found an approach that just adjusts the choice of reference used with the rtol?
IIUC that wouldn't actually mean changing the norms, it would just mean adjusting a couple of lines like this one:

https://github.com/patrick-kidger/lineax/blob/11e95fbec66fe263e4c3e705c12ccd386fec722d/lineax/_solver/bicgstab.py#L96

so that we use something other than `vector` there.

What would you change it to?

---

One point that's not yet clear to me: we speculated that the current approach goes wrong solely due to floating-point issues (since in the infinite-precision limit one expects convergence in a finite number of iterations). Were you able to identify whether that is indeed the case?

jpbrodrick89:
> we speculated that the current approach goes wrong solely due to floating-point issues (since in the infinite-precision limit one expects convergence in a finite number of iterations). Were you able to identify whether that is indeed the case?

So as far as I can tell without having an infinite precision version of JAX, yes, this is indeed the case in general except when `b` is not actually in the range of `A` (in which case iterative solvers can't converge because A is singular). The specific example given above is clearly due to finite precision (occurs due to precision being near round-off floor). Furthermore, if I manually switch to `float32` iterative breakdown can be triggered at much looser tolerances (by the expected ~8 orders of magnitude) or less stark scale differences between coupled blocks.

>  Ah, interesting – you're saying that you've found an approach that just adjusts the choice of reference used with the rtol?

So my "elementwise_rtol" idea would have done a conditional double-gated approach, something like:

```python
b_norm_scale= self.atol + self.rtol * self.norm(vector)
y_norm_scale = self.atol + self.rtol * self.norm(y)
converged = (self.norm(r) > b_norm_scale) | (self.norm(diff) > y_norm_scale)
if not converged:
    return False
if self.elementwise_rtol:
    b_elementwise_scale = (self.atol + self.elementwise_rtol * ω(vector).call(jnp.abs)).ω
    y_elementwise_norm = (self.atol + self.rtol * ω(y).call(jnp.abs)).ω
    norm1 = self.norm((r**ω / b_elementwise_scale**ω).ω)
    norm2 = self.norm((diff**ω / y_elementwise_scale**ω).ω)
    return (norm1 > 1) | (norm2 > 1)
```

> IIUC that wouldn't actually mean changing the norms, it would just mean adjusting a couple of lines like this one:

Actually, instead of using `elementwise_rtol`, my now preferred solution now would now be to introduce an `elementwise_norm` as defined above and just do the following (defaulting to `max_norm` like many other solver libraries):

```python
b_args = (self.atol, self.rtol, vector)
y_args = (self.atol, self.rtol, y)
b_scale = self.atol + self.rtol * self.norm(vector, b_args) # just ones for elementwise_norm
y_scale = self.atol + self.rtol * self.norm(y, y_args) # just ones for elementwise_norm
return (self.norm(r, b_args) > b_scale) | (self.norm(diff, y_args) > y_scale)
```

We don't edit any EXISTING norms other than allow them to accept (but then ignore) a second `args` argument. Noting this would be a breaking change for users using custom norms who would need to add second argument support (I would be very surprised if there are any affected by this).

jpbrodrick89:
@patrick-kidger I'm hoping to get this into the next release but I'm nervous about introducing a breaking change here without your buy-in. What do you think about my solution: switch to standard `max_norm` (which now accepts and ignore a second `args` argument) in accordance with the majority of other solver libraries and offer `elementwise_norm` for backwards compatibility.  I have pushed my proposed solution to this branch so you can see it in action and make an informed judgement.

patrick-kidger:
Ah sorry, I'm clearly holding this up. I think we shouldn't let my mostly-conceptual concerns get in the way of practicality.

If I understand you correctly, then we could fix this by either swapping from `norm(a/b)` to `norm(a)/norm(b)`, or possibly by swapping to `rms_norm` as the default in the linear solvers? If so then I vote that we do one of these two things.

(If at all possible I'd like not to introduce `elementwise_norm` as written here. It's a bit magic – and I'm strongly of the opinion that the default should 'just work', and the extra `args=...` which are only used on some code paths are a bit inscrutable.)

Would that resolve things?

jpbrodrick89:
> If I understand you correctly, then we could fix this by either swapping from `norm(a/b)` to `norm(a)/norm(b)`

This is precisely what my original approach (see [22d4042](https://github.com/patrick-kidger/lineax/pull/230/commits/22d4042efa2e114b56d7ce0ec20cd9802b777d77)) did. I'm perfectly happy to revert my most recent commit back to that (making exact replication of previous behaviour impossible in favour of "just working") if you are.

> or possibly by swapping to rms_norm as the default in the linear solvers?

This doesn't fix anything **on its own**, it is the swap from `norm(a/b)` to `norm(a)/norm(b)` that provides the fix. 

On the default norm itself, the three trade off differently:
- max_norm (current default): zero-pad invariant ✓ and dimension-independent ✓, but not the usual linear-solver convention.
- two_norm: the linear-solver standard (scipy, PETSc) and zero-pad invariant ✓, but dimension-dependent (grows with the number of nonzero entries).
- rms_norm: dimension-independent ✓ but not zero-pad invariant — appending zeros dilutes √mean — so it conflicts with your zero-padding principle.

I'm ambivalent and will gladly adopt whichever you prefer; the norm choice isn't the cause of the issue, only how it's used. My mild lean is to keep max_norm as the only option satisfying both invariances making it closer to the original intended behaviour, but I'm equally happy with two_norm if you'd rather match scipy/PETSc. I'd recommend against `rms_norm`.

patrick-kidger:
Okay, so IIUC then we want `max_norm(a)/max_norm(b)`? Let's do that then :-)

jpbrodrick89:
> Okay, so IIUC then we want max_norm(a)/max_norm(b)? Let's do that then :-)> 

Agreed — went with `max_norm(a) ≤ atol + rtol·max_norm(b)` which is equivalent to your implied `max_norm(a)/max_norm(atol + rtol*b)` when `atol=0` and makes more sense when atol is nonzero. I'll merge now (once tests pass) so we have more robust iterative solvers to play around with in dev before the release, but lmk if I misunderstood or you change your mind about this, I'd be happy to revisit.

PR #231, opened 2026-06-08
Add `project` helper to compute projection operator using Golub-Pereyra

jpbrodrick89:
I'll add more description to thie later two quick things to note:

1. The reason why I've proposed to put this in lineax rather than lineaxtra is due to the use of project in the jvp rule
2. I am beginning to really think it might be worthwhile separating direct and iterative solvers, in that case I would give all direct solvers a project method and huge amount of horrible jankiness can be avoided by replacing with isinstance(DirectLinearSolver)


jpbrodrick89:
Shelving this until slogdet is in. I'm happy with the public API and primitive here but I don't like the way I've done the project fast path and have some other ideas that depend on adding AbstractDirectLinearSolver.

PR #233, opened 2026-06-18
refactor: move Abstract/AutoLinearOperator to separate files to avoid circular imports

jpbrodrick89:
@patrick-kidger the current setup has been causing me headaches trying to come up with more elegant designs for the jvp fast paths I tried out in #217 (such as creating private type aliases for subsets of solvers in _solve.py for type hinting and isinstance checking that depend on having access to solvers at module level). Moving AbstractLinearOperator solves a lot as it depends on nothing. We now have the much cleaner dependency tree

AbstractLinearOperator (now in _solver/base.py) -> concrete solvers (in _solver/) -> AutoLinearSolver (now in _solver/auto.py) -> linear_solve(_p) (still in _solve.py)

Is there any good reason not to move AbstractLinearOperator to its own file? It seems to clean up a lot. There is now just one circular import in GMRES which uses linear_solve internally which is generally a case worth special treatment (and actually one on the laundry list of moderate inefficiencies to address).

Moving AutoLinearSolver doesn't actually solve any problem, I just couldn't see a reason not to do it and I like the conceptual neatness of having all solvers in `_solver/`

### DIFF
--- a/lineax/_solve.py
+++ b/lineax/_solve.py
@@ -342,8 +342,6 @@ def _check_rank_compat(
             )
 
 
-# TODO(kidger): gmres, bicgstab
-# TODO(kidger): support auxiliary outputs
 @eqx.filter_jit
 def linear_solve(
     operator: AbstractLinearOperator,
--- a/lineax/_solve.py
+++ b/lineax/_solve.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 import functools as ft
-from typing import Any, TypeAlias
+from typing import Any
 
 import equinox as eqx
 import equinox.internal as eqxi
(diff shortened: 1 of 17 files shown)


lineax/_solve.py
```diff
@@ -490,177 +342,6 @@ def _check_rank_compat(
...
-
-
-AutoLinearSolver.__init__.__doc__ = """**Arguments:**
-
-- `well_posed`: whether to only handle well-posed systems or not, as discussed above.
-"""
-
-
 # TODO(kidger): gmres, bicgstab
 # TODO(kidger): support auxiliary outputs
```
jpbrodrick89: while we're here, do you know what these TODO's were referring to? We have GMRES and BiCGSTAB and I think we've explicitly decided to never support auxiliary outputs (e.g. by deprecating `AuxLinearOperator`) so can we remove?

patrick-kidger:
No strong feelings! Seems reasonable to me :)

lineax/_solve.py
patrick-kidger: Yup I think so!

jpbrodrick89:
```suggestion
```

PR #237, opened 2026-07-07
fix(QR): scale operator before Householder factorization to avoid overflow

### LINKED ISSUE #98: Bug in QR solver for Tikhnov regularised matrices
When solving against matrices of the form $A + \lambda I$ with $\lambda$ set to the max dtype value, QR solves return `nan` instead of zero. This does not occur with matrices of the form $\lambda I$ and for some matrices $A$ it seems not to be an issue, but for dense $A$ it is a problem.

This case arises in the Levenberg-Marquardt algorithm in [Optimistix](https://github.com/patrick-kidger/optimistix) when the step-size goes to zero.

This failure is not silent, but it should be supported.

MWE:
```python
import jax
import jax.numpy as jnp
import jax.random as jr
import lineax as lx

vector = jnp.ones(100)
# does not need to be a `MatrixLinearOperator`.
dense_operator = lx.MatrixLinearOperator(0.1 * jr.normal(jr.PRNGKey(0), (100, 100)))
dampening = jnp.finfo(jnp.asarray(0.0)).max
tikhnov_operator = dense_operator + dampening * lx.IdentityLinearOperator(
    jax.eval_shape(lambda: vector)
)
linear_sol = lx.linear_solve(tikhnov_operator, vector, solver=lx.QR(), throw=False)

assert not jnp.any(jnp.isnan(linear_sol.value))  # assertion fails
```

Sumu004:
## What

`lx.QR()` returns `nan` for matrices of the form `A + lambda*I` when `lambda` is near the max representable dtype value, instead of the correct (near-zero) result:

```python
import jax
import jax.numpy as jnp
import jax.random as jr
import lineax as lx

vector = jnp.ones(100)
dense_operator = lx.MatrixLinearOperator(0.1 * jr.normal(jr.PRNGKey(0), (100, 100)))
dampening = jnp.finfo(jnp.asarray(0.0)).max
tikhnov_operator = dense_operator + dampening * lx.IdentityLinearOperator(
    jax.eval_shape(lambda: vector)
)
linear_sol = lx.linear_solve(tikhnov_operator, vector, solver=lx.QR(), throw=False)
assert not jnp.any(jnp.isnan(linear_sol.value))  # assertion fails on main
```

This arises in Optimistix's Levenberg-Marquardt as the step size goes to zero.

## Root cause

Householder QR computes reflector norms via a sum-of-squares formula. Once an entry of the input matrix approaches `sqrt(dtype_max)`, squaring it inside that norm computation overflows to `inf`, which then propagates into **both** `Q` and `R` — I confirmed this directly (not just at the final solve step): for a matrix with `lambda = float64 max` on the diagonal, `Q` itself already contains `inf`/`nan` before any triangular solve is attempted. This isn't unique to JAX's QR implementation; it follows from computing a Euclidean norm the naive way (`sqrt(sum(x**2))`) for any input whose square exceeds the dtype's representable range — the same failure mode affects LAPACK-based QR in plain NumPy too, which I used to confirm the mechanism without needing to touch JAX.

## Fix

Scale the matrix by its own max-abs entry before calling `jnp.linalg.qr`, keeping every value passed to the Householder factorization within `[~0, 1]` regardless of the original matrix's dynamic range. This doesn't change the mathematical solution — `A x = b` has the same solution as `(A/s) x = b/s` for any nonzero scalar `s` — so `compute` divides the right-hand side by the same `scale` rather than multiplying `R` back up, which keeps every intermediate value in the well-conditioned regime established in `init` instead of reintroducing the overflow risk we just scaled away from. Applied symmetrically to both the standard and transpose (underdetermined) solve branches, since both go through the same Householder factorization and are subject to the identical overflow mechanism.

## Verification

Checked against plain NumPy (confirming the underlying floating-point mechanism independent of JAX):
- The scaled formulation matches the expected `vector / lambda` result to full float64 precision for the reported repro.
- Matches a direct solve to ~1e-15 for well-conditioned matrices — i.e. the fix is a no-op for normal-scale operators, as expected since scaling by a nonzero constant doesn't change the solution.
- Same holds for the transpose/underdetermined branch, tested on a rectangular Tikhonov-damped operator.

## Checklist

- [x] New tests in `tests/test_solve.py`:
  - `test_qr_tikhonov_extreme_dampening`: the exact reported repro
  - `test_qr_matches_direct_solve_normal_case`: confirms the fix is a no-op for well-conditioned operators
  - `test_qr_underdetermined_extreme_dampening`: same overflow scenario on the transpose (underdetermined) branch
- [x] `black` formatting clean
- [x] Existing `QR` tests (mixed dtypes, complex, triangular) unaffected — the scaling is transparent for well-scaled operators

Fixes #98

### DIFF
--- a/lineax/_solver/qr.py
+++ b/lineax/_solver/qr.py
@@ -30,8 +30,7 @@ from .misc import (
     unravel_solution,
 )
 
-
-_QRState: TypeAlias = tuple[tuple[Array, Array], eqxi.Static, PackedStructures]
+_QRState: TypeAlias = tuple[tuple[Array, Array, Array], eqxi.Static, PackedStructures]
 
 
 class QR(AbstractLinearSolver):
@@ -59,10 +58,21 @@ class QR(AbstractLinearSolver):
         transpose = n > m
         if transpose:
             matrix = matrix.T
-        h, taus = jnp.linalg.qr(matrix, mode="raw")  # pyright: ignore
+        # Householder QR computes reflector norms via `sqrt(sum(x**2))`, which
+        # overflows to `inf` (and subsequently `nan`) whenever an entry of
+        # `matrix` approaches `sqrt(dtype_max)` -- e.g. a Tikhonov-damped
+        # operator `A + lambda * I` with `lambda` near the max representable
+        # value. Scaling by the matrix's own max-abs entry keeps every value
+        # passed to `jnp.linalg.qr` safely within [~0, 1], which avoids the
+        # overflow without changing the result for well-scaled operators
+        # (scaling a linear system by a nonzero constant doesn't change its
+        # solution). `scale` is corrected back out in `compute`.
+        scale = jnp.max(jnp.abs(matrix))
+        scale = jnp.where(scale == 0, jnp.ones_like(scale), scale)
+        h, taus = jnp.linalg.qr(matrix / scale, mode="raw")  # pyright: ignore
         a = h.mT
         packed_structures = pack_structures(operator)
-        return (a, taus), eqxi.Static(transpose), packed_structures
+        return (a, taus, scale), eqxi.Static(transpose), packed_structures
 
     def compute(
         self,
@@ -70,17 +80,24 @@ class QR(AbstractLinearSolver):
         vector: PyTree[Array],
         options: dict[str, Any],
     ) -> tuple[PyTree[Array], RESULTS, dict[str, Any]]:
-        (a, taus), transpose, packed_structures = state
+        (a, taus, scale), transpose, packed_structures = state
         transpose = transpose.value
         del state, options
         vector = ravel_vector(vector, packed_structures)
         n_full, n_min = a.shape
         r = a[:n_min]
+        # `a` (and thus `r`) was computed from `matrix / scale`, so `r` is R
+        # for the *scaled* operator. Dividing the right-hand side by `scale`
+        # here (rather than multiplying `r` back up) keeps every intermediate
+        # value in the well-scaled regime established in `init`, instead of
+        # reintroducing the same overflow risk we scaled away from.
         if transpose:
             # Minimal norm solution if underdetermined: x = Q.conj() @ R^{-T} @ b.
             # Use Q.conj() @ z = (z^T @ Q^H)^T to avoid explicit `conj` calls,
             # and pad `y` along the row axis to absorb the discarded columns of Q.
-            y = jsp.linalg.solve_triangular(r, vector, trans="T", unit_diagonal=False)
+            y = jsp.linalg.solve_triangular(
+                r, vector / scale, trans="T", unit_diagonal=False
+            )
             zeros = jnp.zeros((1, n_full - n_min), dtype=y.dtype)
             y_pad = jnp.concatenate([y[None, :], zeros], axis=1)
             solution = jll.ormqr(a, taus, y_pad, left=False, transpose=True)[0]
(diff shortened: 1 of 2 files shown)


jpbrodrick89:
Hi @Sumu004, thanks for giving this a spin! Yes, the MWE in #98 does still create `nan`'s so the "bug" is still present. However, this is just inherent behaviour of `jnp.linalg.qr` and not something I feel we should fix INSIDE the solver itself. Moreover, the referenced `LevenbergMarquadt` path in the issue either erroneous or historical, as far as I can tell `QR` is never use to directly solve `A + lambda*I` instead it solves `[J, sqrt(lambda*I)]` (the Jacobian appended with Tikhonov regularisation rows) which doesn't have this problem at the same `lambda`. QR should rarely be used for square matrices so #98 is very unlikely to trigger.

If you do need a workaround I would suggest a `ScaledSolver` wrapper mirroring `Normal` so it can be used with any solver rather than special-casing QR, if there are more use cases I might entertain having such a wrapper in `lineax` proper but I'm not convinced the justification is quite there yet:

```python
from typing import Any, TypeVar
import jax.numpy as jnp
import jax.tree_util as jtu
import lineax as lx

_InnerState = TypeVar("_InnerState")


class ScaledSolver(lx.AbstractLinearSolver):
    """Wraps any inner solver, rescaling the operator by its max-abs entry
    before factorisation so no intermediate exceeds ~1 in magnitude.

    Solves `A x = b` by handing the inner solver the equivalent system
    `(A / s) x = (b / s)`, where `s = max_ij |A_ij|`. Since `s` is a nonzero
    scalar the solution is unchanged, but every value the factorisation sees is
    now O(1) -- avoiding the `sqrt(sum(x**2))` overflow that makes e.g. `QR`
    return `nan` for operators with entries above `sqrt(dtype_max)`.
    """

    inner_solver: lx.AbstractLinearSolver

    def init(self, operator, options):
        scale = jnp.max(jnp.abs(operator.as_matrix()))
        scale = jnp.where(scale == 0, jnp.ones_like(scale), scale)
        inner_state = self.inner_solver.init(operator / scale, options)
        return inner_state, scale

    def compute(self, state, vector, options):
        inner_state, scale = state
        vector = jtu.tree_map(lambda v: v / scale, vector)
        return self.inner_solver.compute(inner_state, vector, options)

    def transpose(self, state, options):
        inner_state, scale = state
        inner_state_t, options_t = self.inner_solver.transpose(inner_state, options)
        return (inner_state_t, scale), options_t

    def conj(self, state, options):
        inner_state, scale = state  # scale is real, unaffected by conjugation
        inner_state_c, options_c = self.inner_solver.conj(inner_state, options)
        return (inner_state_c, scale), options_c

    def assume_full_rank(self):
        return self.inner_solver.assume_full_rank()
```

Usage: lx.linear_solve(operator, b, solver=ScaledSolver(lx.QR())).

Sumu004:
Thanks for the detailed correction, and for taking the time to trace through the actual LM code path.

You're right about the mechanism. I went back and re-checked `damped_newton_step`: the least-squares branch builds the augmented operator `[J; sqrt(lambda)*I]` via `_Damped`, so the matrix entries scale with `sqrt(lambda)`, not `lambda` directly. Squaring `sqrt(lambda)` inside the Householder norm recovers `lambda` itself, not `lambda^2`, so the overflow threshold in practice sits right at `lambda ~ dtype_max`, not below it as I'd assumed from the issue's square-matrix repro. So the original issue's synthetic `dense_operator + dampening*IdentityLinearOperator` case (a square matrix, dampening at the literal dtype max) isn't really representative of how `QR` gets exercised inside LM, and your point that `QR` on square matrices is a rare path generally stands too.

I agree the fix doesn't belong baked into `QR` itself given that. Your `ScaledSolver` wrapper is a cleaner shape for this than what I had (opt-in, works with any solver, doesn't touch the common case at all). Since you've already written it out: would you like me to open a fresh PR with that implementation (crediting the design to you in the description), or would you rather add it yourself on your own timeline? Either way works for me, and I'm happy to close this PR now given the original scoping doesn't hold up.

jpbrodrick89:
Thanks for your understand @Sumu004. My gut feel is that this issue is fairly niche and the scaled solver would see insufficient usage to warrant supporting it especially if it advertises more numerical robustness than we can guarantee (although maybe putting it in a FAQ is an option). Have you run into this issue yourself in your work? @packquickly @patrick-kidger do you have any lean here?

Sumu004:
Honestly, no, I haven't hit this in my own work. I found it while going through open issues looking for real numerical bugs to work on, not from a project that needed it. So I don't have a usage-frequency data point to offer here, and I think that's relevant context for your call given the "insufficient usage" concern.

Given that, I'll go ahead and close this PR now rather than wait on a lean from packquickly/patrick-kidger for something I can't actually vouch for the need of. Thanks for engaging with it either way, appreciate you taking the time to trace through the LM code path.

PR #238, opened 2026-07-28
Circulant operator

BalzaniEdoardo:
Thanks for the detailed review — I should have addressed all of the inline comments. 

## Summary

Adds support for circulant matrices: an operator that stores only $O(n)$ data, and a solver that runs in $O(n \log n)$ instead of $O(n^3)$.

A circulant matrix is fully determined by its first column $c$, via $A_{ij} = c_{(i - j) \bmod n}$. It is diagonalised by the DFT, so its eigenvalues are just $\mathrm{FFT}(c)$ — which makes both the matrix-vector product and the solve a pair of FFTs and a pointwise multiply or divide.

## Implementation

- `CirculantLinearOperator` — stores the first column; `mv` via `rfft`/`irfft`, `as_matrix` via `jax.scipy.linalg.circulant`.
- `Circulant` — divides by the eigenvalues. `well_posed=False` (the default) zeroes modes below an `rcond` cutoff, giving the pseudoinverse solution.
- `circulant_tag`, `is_circulant`, and `first_column`, the last completing the `diagonal` / `tridiagonal` family.
- `AutoLinearSolver` dispatch, and `try_structured_materialise` returning a `CirculantLinearOperator` for a tagged operator with flat structures.
- `is_circulant` propagates through addition and, unusually, through composition too — all circulant matrices share the DFT eigenbasis, so the product of two is circulant.

## Things to flag

Three things came out of the changes that are worth flagging explicitly.

### `first_column` defaulting to a matvec introduced a recursion

Combining the two suggestions — have `try_structured_materialise` return a `CirculantLinearOperator` via `first_column`, and let `first_column` fall back to `operator.mv(e_0)` for unregistered types — closes a loop:

```
try_structured_materialise → first_column → mv → try_structured_materialise → ...
```

Fixed by registering `first_column` for `MatrixLinearOperator`, `PyTreeLinearOperator`, `AddLinearOperator` and `ComposedLinearOperator`. Worth knowing that the matvec default is only safe once those concrete types are covered.

### `test_jvp` reference changed from `lstsq` to `solve`

The circulant tag was producing seed-dependent JVP failures (2, 2 and 4 failures across three runs). The cause is not the operator: a **real** circulant matrix has exactly degenerate singular values. SVD derivatives are undefined at repeated singular values, so the `jnp.linalg.lstsq`returned NaN sometimes, or was silently wrong otherwise.

So the reference now uses `solve` rather than `lstsq` when `not pseudoinverse`.

### Squareness in the `is_*` checks

On whether `is_circulant(IdentityLinearOperator)` should check squareness: `structured.py` currently splits into two groups. The square-only (`is_symmetric`, `is_positive_semidefinite`) and the shape-agnostic ones (`is_diagonal`, `is_tridiagonal`, both triangular checks, unit-diagonal). Circulant is written with the first group. Both solvers raise on non-square input regardless, so this is about which convention I should follow with the implementation. Happy to move it either way.

### Dependency floor

`as_matrix` now uses `jax.scipy.linalg.circulant` (instead of the original `jnp.ogrid` trick), which was added in **jax 0.10.1**. The `jax>=0.10.0` floor in `pyproject.toml` needs bumping to `>=0.10.1` accordingly if we want to keep the current implementation, otherwise we can revert to the `ogrid`.


lineax/_operator/binary.py
```diff
@@ -229,11 +231,14 @@ def _(operator):
     )
 
 
-# These properties ARE preserved under composition
+# These properties ARE preserved under composition. (Circulant matrices are all
+# simultaneously diagonalised by the DFT, so they commute and their product is again
+# circulant.)
```
jpbrodrick89:
```suggestion
# These properties ARE preserved under composition.
```

No need for special comment on circulant matrices here (circulant matrices forming a closed algebra is a well known fact).

lineax/_operator/core.py
jpbrodrick89: Incorporate into the loop below.

lineax/_operator/structured.py
```diff
@@ -245,11 +273,58 @@ def out_structure(self):
...
+        else:
+            freq_circulant = jnp.fft.rfft(self.column)
+            freq_vector = jnp.fft.rfft(vector)
+            (size,) = self.column.shape
+            result = jnp.fft.irfft(freq_circulant * freq_vector, n=size)
+        return result
+
+    def transpose(self):
+        return CirculantLinearOperator(
+            jnp.concatenate([self.column[:1], self.column[1:][::-1]]),
```
jpbrodrick89: Use `jnp.flip` like in the solver.

lineax/_operator/structured.py
```diff
@@ -245,11 +273,58 @@ def out_structure(self):
...
+            transpose_tags(self.tags),
+        )
+
+    def as_matrix(self):
+        (size,) = jnp.shape(self.column)
+        # static indices, use numpy
+        i, j = np.ogrid[:size, :size]
+        return self.column[(i - j) % size]
```
jpbrodrick89: any reason not to use [jax.scipy.linalg.toeplitz](https://docs.jax.dev/en/latest/_autosummary/jax.scipy.linalg.toeplitz.html) here?

lineax/_operator/structured.py
```diff
@@ -329,13 +452,56 @@ def _(operator):
...
+        operator.in_size() < 3
+        or tridiagonal_tag in operator.tags
+        or diagonal_tag in operator.tags
+    )
+
+
+@is_circulant.register(IdentityLinearOperator)
+def _(operator):
+    # A non-square `IdentityLinearOperator` is not circulant.
+    return eqx.tree_equal(operator.in_structure(), operator.out_structure()) is True
```
jpbrodrick89, line 438: I'm not sure about this as it is inconsistent with `@is_tridiagonal.register(IdentityLinearOperator)`: `is_tridiagonal` typically requires "square" matrices (i.e. `in_structure == out_structure`) but does not check for this, either that's a bug and needs to be more strict or this needs to be more loose.

lineax/_operator/structured.py
```diff
@@ -35,13 +35,27 @@
...
+    diagonal_tag,
+    lower_triangular_tag,
+    negative_semidefinite_tag,
+    positive_semidefinite_tag,
+    symmetric_tag,
+    transpose_tags,
+    tridiagonal_tag,
+    unit_diagonal_tag,
+    upper_triangular_tag,
+)
```
jpbrodrick89: I'm in two minds about whether to do this and treat Circulant Linear Operator as an exceptional structured operator that accepts tags or whether to have the same contract as other structured operators and just expect users to use `TaggedLinearOperator`, what are the primary reasons for going with this design?

lineax/_operator/base.py
```diff
@@ -399,6 +399,48 @@ def tridiagonal(
...
+    # third-party subclasses that do not register a dispatch still work.
+    in_structure = operator.in_structure()
+    if (
+        not isinstance(in_structure, jax.ShapeDtypeStruct)
+        or len(in_structure.shape) != 1
+    ):
+        # Circulance is not defined for PyTree-structured spaces.
+        _default_not_implemented("circulant_column", operator)
+    (size,) = in_structure.shape
+    basis = jnp.zeros(size, in_structure.dtype).at[0].set(1)
```
jpbrodrick89: use `with jax.ensure_compile_time_eval` in line with diagonal basis and tridiagonal basis in `try_structured_materialise`

lineax/_operator/base.py
```diff
@@ -399,6 +399,48 @@ def tridiagonal(
...
+    # raise `NotImplementedError` for unregistered types: one matvec against `e_0` is a
+    # correct (if unoptimised) answer for any operator over a flat vector space, so
+    # third-party subclasses that do not register a dispatch still work.
+    in_structure = operator.in_structure()
+    if (
+        not isinstance(in_structure, jax.ShapeDtypeStruct)
+        or len(in_structure.shape) != 1
+    ):
+        # Circulance is not defined for PyTree-structured spaces.
+        _default_not_implemented("circulant_column", operator)
```
jpbrodrick89: I'd reconsider this, look at how `try_structured_materialise` extracts the diagonal for diagonal Jacobian/Function linear operators by ravelling, are you able to use that approach here?

lineax/_operator/base.py
```diff
@@ -399,6 +399,48 @@ def tridiagonal(
...
+def circulant_column(operator: AbstractLinearOperator) -> Shaped[Array, " size"]:
+    """Extracts the first column from a circulant linear operator, and returns a
+    vector.
+
+    A circulant matrix is completely determined by its first column `c`, via
+    `matrix[i, j] = c[(i - j) % size]`, so this is the circulant analogue of
+    [`lineax.diagonal`][] and [`lineax.tridiagonal`][].
+
+    This is only meaningful for operators that are actually circulant; use
+    [`lineax.is_circulant`][] to check.
```
jpbrodrick89: Why do you not check for this inside `circulant_column`? If its because you want to reuse this for general Toeplitz I would consider separating out the logic. We could have a general "first column" helper that works for any matrix and then `circulant_column` just calls this as a default after running any checks. Alternatively (and potentially preferably) you could just RENAME this function as `first_column`, the algorithm you use here actually makes no assumption circulancy so that's probably the right approach here.

lineax/_operator/base.py
```diff
@@ -399,6 +399,48 @@ def tridiagonal(
     _default_not_implemented("tridiagonal", operator)
 
 
+@ft.singledispatch
+def circulant_column(operator: AbstractLinearOperator) -> Shaped[Array, " size"]:
```
jpbrodrick89: I would be tempted to call this within `try_structured_materialise` to return a `CirculantLinearOperator`.

lineax/_solver/circulant.py
```diff
@@ -0,0 +1,109 @@
...
+            conj_freq = jnp.conjugate(column_fft)
+            conj_freq = jnp.concatenate([conj_freq[:1], jnp.flip(conj_freq[1:])])
+            conj_state = ((conj_freq, is_complex, n), packed_structures)
+        else:
+            conj_state = state
+        conj_options = {}
+        return conj_state, conj_options
+
+    def assume_full_rank(self):
+        return True
```
jpbrodrick89: I think we should offer a non full rank option too with allowing `well_posed` argument like `Diagonal(well_posed=...)` and then using where to filter out `< rcond` values and instead multiply by zero rather than divide by a very small number in the same way.

lineax/_solver/circulant.py
```diff
@@ -0,0 +1,109 @@
...
+                "`Circulant` may only be used for linear solves with square matrices"
+            )
+        if not is_circulant(operator):
+            raise ValueError(
+                "`Circulant` may only be used for linear solves with circulant matrices"
+            )
+        column = circulant_column(operator)
+        is_complex = jnp.iscomplexobj(column)
+        if is_complex:
+            fft_column = jnp.fft.fft(column)
```
jpbrodrick89: `fft_column` is just the eigenvalues right? Lets call it eigenvalues then.

lineax/_solver/circulant.py
```diff
@@ -0,0 +1,109 @@
...
+        else:
+            fft_column = jnp.fft.rfft(column)
+        return (fft_column, is_complex, len(column)), pack_structures(operator)
```
jpbrodrick89: There's no point in putting static information such as `is_complex` and `len(column)` in state.

lineax/_solver/circulant.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def transpose(self, state: _CirculantState, options: dict[str, Any]):
+        del options
+        (column_fft, is_complex, n), packed_structures = state
+        transposed_packed_structures = transpose_packed_structures(packed_structures)
+        # Transposing reverses the column, `c[(-k) % n]`, and reversal negates the
+        # frequency index: `λ_k -> λ_{-k}`. `rfft` keeps only half the spectrum, on
+        # which that reindexing acts as conjugation.
+        if is_complex:
+            transpose_freq = jnp.concatenate([column_fft[:1], jnp.flip(column_fft[1:])])
```
jpbrodrick89: this is the same operation as transposing the column, I'm wondering whether its worth adding a helper function? If you think that's overkill (it probably is!) no worries.

lineax/_solver/circulant.py
```diff
@@ -0,0 +1,109 @@
...
+    ) -> tuple[PyTree[Array], RESULTS, dict[str, Any]]:
+        (fft_column, is_complex, n), packed_structures = state
+        del state, options
+        vector = ravel_vector(vector, packed_structures)
+        if is_complex:
+            fft_fn = jnp.fft.fft
+            ifft_fn = jnp.fft.ifft
+        else:
+            fft_fn = jnp.fft.rfft
+            ifft_fn = lambda x: jnp.fft.irfft(x, n=n)
```
jpbrodrick89: use partial here rather than lambda

lineax/_operator/structured.py
```diff
@@ -245,11 +273,58 @@ def out_structure(self):
...
+        tags: object | frozenset[object] = (),
+    ):
+        self.column = inexact_asarray(column)
+        if self.column.ndim != 1:
+            raise ValueError("Circulant must have exactly 1 dimension.")
+        self.tags = as_frozenset(tags)
+
+    def mv(self, vector):
+        if jnp.issubdtype(self.column.dtype, jnp.complexfloating):
+            freq_circulant = jnp.fft.fft(self.column)
```
jpbrodrick89: is it worth calling these eigenvalues or not so meaningful here?

lineax/_operator/structured.py
```diff
@@ -291,6 +371,30 @@ def _(operator):
...
+        lower_diag = jnp.full(diag.size - 1, operator.column[1], dtype=diag.dtype)
+    return diag, lower_diag, upper_diag
+
+
+@circulant_column.register(IdentityLinearOperator)
+def _(operator):
+    size = operator.in_size()
+    dtype = jtu.tree_leaves(operator.in_structure())[0].dtype
+    return jnp.zeros(size, dtype).at[0].set(1)
+
```
jpbrodrick89: This is a lot more awkward if we decide to handle PyTree operators but actually for IdentityLinearOperator I think the default implementation of `circulant_column` is no faster anyway (`IdentityLinearOperator.mv` is a runtime no-op, all work is done at compile time).

lineax/_operator/structured.py
```diff
@@ -291,6 +371,30 @@ def _(operator):
     return operator.diagonal, operator.lower_diagonal, operator.upper_diagonal
 
 
+@tridiagonal.register(CirculantLinearOperator)
+def _(operator):
+    diag = diagonal(operator)
+    if diag.size == 1:
+        upper_diag = jnp.zeros(0, dtype=diag.dtype)
+        lower_diag = jnp.zeros(0, dtype=diag.dtype)
```
jpbrodrick89: These are just empty arrays right? A case that would never be used but `jnp.array([],...)` might be more informative (matter of taste, and I don't feel strongly about it).

tests/test_operator.py
```diff
@@ -238,6 +252,17 @@ def test_tridiagonal(dtype, getkey):
     assert jnp.allclose(upper_diag, jnp.diagonal(td_matrix, 1))
 
 
+@pytest.mark.parametrize("dtype", (jnp.float64, jnp.complex128))
+def test_circulant(dtype, getkey):
```
jpbrodrick89: this is just `test_first_column` the matrix doesn't have to be circulant for it to work.

jpbrodrick89 (changes requested):
Nice work @BalzaniEdoardo! Despite still being under draft, this is looking pretty similar to what I had in mind so thanks a great first PR! I'd really like to get the rank-deficient circulant solver included in this PR and it should be fairly straightforward. Happy to discuss any of the comments further.

You've probably realised this already, but it might be worth stating in the comments that its IMPOSSIBLE for a matrix to be simultaneously tridagonal and circulant as I forgot this fact and got a bit tied up thinking about prioritisation and then realised it didn't matter.

I believe there is some missing test coverage, consider sites like:

1. tests/helpers.py:157 params() — no skip clause for make_circulant_operator, unlike the three that exist for the diagonal/identity/tridiagonal makers. It gets paired with every tag in the table and hits assert False, tags at helpers.py:329. Breaks test_transpose.py:41, test_well_posed.py:31, test_singular.py:34.
2. tests/helpers.py:114 solvers_tags_pseudoinverse — no (lx.Circulant(), lx.circulant_tag, False) row, so even with (1) fixed the circulant operator is never exercised by those suites and lx.Circulant gets zero solver coverage (and we will definitely want this when we have a rank deficient circulant solver).
3. tests/helpers.py:31 _construct_matrix_impl — no circulant_tag branch alongside the existing diagonal_tag/tridiagonal_tag ones. This is the subtle one: with (1) and (2) fixed but not this, construct_matrix returns a random matrix, make_circulant_operator takes matrix[:, 0], and the operator silently represents a different matrix than the one the test compares against — a wrong-answer failure, not an assert. The branch should go early (before symmetric_tag), and the existing while True conditioning loop will reject ill-conditioned draws.
4. tests/test_adjoint.py:20 — needs a make_circulant_operator branch like the tridiagonal one at :27-31; it currently falls through to else and gets a non-square (3, 5) matrix with tags=(), hitting the same assert False.

Thanks again!

lineax/_operator/structured.py
BalzaniEdoardo: removed

lineax/_operator/structured.py
BalzaniEdoardo: done a bit of digging, actually using numpy there is a bad idea. The `np.ogrid` are static and hardcoded in the compiled code. Compile time and size of the compiled program grows with n.

Toepliz is not a jax primitive, is concatenate + lax.conv_general_dilated_patches + two flips.

The most efficient approach is to use `jnp.ogrid` instead, this doesn't have the same scalability issue of the numpy counterpart. The toepliz route scales poorly with the matrix size.

Below a little benchmarking code run: 

**Setup:**

`as_matrix` candidate implementations.

```python
>>> import jax, timeit
>>> jax.config.update("jax_enable_x64", True)
>>> import jax.numpy as jnp, numpy as np, jax.scipy.linalg as jsl
>>> def np_grid(c):
...     (n,) = c.shape; i, j = np.ogrid[:n, :n]; return c[(i - j) % n]
>>> def jnp_grid(c):
...     (n,) = c.shape; i, j = jnp.ogrid[:n, :n]; return c[(i - j) % n]
>>> def toeplitz(c):
...     return jsl.toeplitz(c, jnp.concatenate([c[:1], jnp.flip(c[1:])]))
```

**Size of constants in jaxpr (MB):**

```python
>>> def consts(f, c):
...     return sum(np.asarray(x).nbytes for x in jax.make_jaxpr(f)(c).consts)
>>> print(f"{'n':>5s} {'np.ogrid':>9s} {'jnp.ogrid':>9s}")
>>> for n in (512, 2048, 8192):
...     c = jnp.zeros(n)
...     print(f"{n:5d} {consts(np_grid, c) / 1e6:9.2f} {consts(jnp_grid, c) / 1e6:9.2f}")
    n  np.ogrid jnp.ogrid
  512      2.36      0.00
 2048     37.75      0.00
 8192    603.98      0.00
```

**Runtime per call (ms):**


```python
>>> def ms(f, c):
...     g = jax.jit(f).lower(c).compile(); g(c).block_until_ready()
...     return min(timeit.repeat(lambda: g(c).block_until_ready(), number=3, repeat=2)) / 3 * 1e3
>>> print(f"{'n':>5s} {'np.ogrid':>9s} {'jnp.ogrid':>9s} {'toeplitz':>11s}")
>>> for n in (64, 256, 1024):
...     c = jnp.zeros(n)
...     print(f"{n:5d} {ms(np_grid, c):9.4f} {ms(jnp_grid, c):9.4f} {ms(toeplitz, c):11.4f}")
    n  np.ogrid jnp.ogrid    toeplitz
   64    0.0176    0.0632      0.2665
  256    0.0532    0.0855      5.9448
 1024    0.4001    0.3938   1029.1881
```

The results above are on my CPU, on cuda it's the same

```
[CudaDevice(id=0)]
jaxpr constants (MB)
    n  np.ogrid jnp.ogrid
  512      2.10      0.01
 2048     33.55      0.03
 8192    536.87      0.13

runtime per call (ms)
    n  np.ogrid jnp.ogrid    toeplitz
   64    0.0338    0.0438      0.0947
  256    0.0355    0.0423      0.1301
 1024    0.0959    0.0379      1.7797
```

jpbrodrick89: I will try look in more detail tomorrow, this is just a first thought that pops into my head reponse: how does jax.scipy.linalg.**circulant** fare? If that is not competitive do you think we should use it for maintainability and do a PR to jax so that jax does the optinal version?

BalzaniEdoardo:
Hi Jonathan!

I did not look into that yet, I had no idea it existed. I was just about to
post a long comment: currently, the fastest method that has no memory issue
is `jnp.ogrid`. I'll look into circulant directly

I'll still post in the PR as refrerence

Il giorno lun 3 ago 2026 alle ore 15:45 Jonathan Brodrick <
***@***.***> ha scritto:

lineax/_operator/structured.py
BalzaniEdoardo: just looked at the source for`jax.scipy.linalg.circulant`, it is the same trick 

```
@partial(jnp_vectorize.vectorize, signature="(n)->(n,n)")
def _circulant(c: Array) -> Array:
  n, = c.shape
  idx = (jnp.arange(n)[:, None] - jnp.arange(n)[None, :]) % n
  return c[idx]
```

BalzaniEdoardo: here is a benchmark:
```
jaxpr constants (MB)
    n      np.ogrid     jnp.ogrid jsl.circulant  jsl.toeplitz
  512          2.36          0.00          0.00          0.00
 2048         37.75          0.00          0.00          0.00
 8192        603.98          0.00          0.00          0.00

runtime per call (ms)
    n      np.ogrid     jnp.ogrid jsl.circulant  jsl.toeplitz
   64        0.0168        0.0865        0.0659        0.1259
  256        0.0935        0.0839        0.0650        6.4513
 1024        0.4864        0.2290        0.3515     1062.5349
```

I am ambivalent, let me know which solution you prefer!

jpbrodrick89: Ok that seems a lot closer in performance (less than a factor of two rather than orders of magnitude), I think we just use jax.scipy.linalg.circulant for now and address the performance issue upstream in jax. The jax guys are usually fast to respond on small well benchmarked noncontroversial optimisations which I think this will qualify as.

BalzaniEdoardo:
@jpbrodrick89 I just realized that a circulant real matrix has pair of equal singular values. The `lstsq` doesn't work well, either returns nans or wrong values. I fixed the `test_jvp` by switching between least-square and solve. Let me know if that's a good solution

lineax/_solver/circulant.py
BalzaniEdoardo: I did made `is_complex` static, and I realize we don't need to store `len(column)`. Let me know if you like the solution

lineax/_solver/circulant.py
BalzaniEdoardo: yes, I renamed everywhere (I believe)

lineax/_solver/circulant.py
BalzaniEdoardo: Added!

lineax/_operator/structured.py
BalzaniEdoardo: I dropped the tags. I did not realize that all other structured operator did not accept the tag. It makes sense to keep the same interface for Circulant

lineax/_solver/circulant.py
BalzaniEdoardo: I ended up adding one

BalzaniEdoardo:
@jpbrodrick89 I added the doc pages and better test coverage. In the description I flagged some decisions that definitely needs a review, in particular raising the jax floor for the `jax.scipy.linalg.circulant`.

lineax/_operator/structured.py
jpbrodrick89: Lets leave for now, I'd prefer to return `True` unconditionally but `IdentityLinearOperator` does not work like I thought, I'm considering reworking `IdentityLinearOperator` to simplify but I need to discuss with Patrick first (as such I will modify this in that follow up PR when I get to it).

lineax/_operator/core.py
jpbrodrick89:
```suggestion
   Returns a structured operator
   (`DiagonalLinearOperator`/`TridiagonalLinearOperator`/`CirculantLinearOperator`)
   if the operator is known to have the required structure (e.g through tags),
   otherwise returns the original operator unchanged. The resulting operator
   preserves the input/output structure of the original operator.
```

lineax/_operator/binary.py
```diff
@@ -201,6 +208,12 @@ def _(operator):
     return main_diagonal, lower_diagonal, upper_diagonal
 
 
+@first_column.register(ComposedLinearOperator)
+def _(operator):
+    # The first column of `A @ B` is `A @ (B e_0)`.
+    return operator.operator1.mv(first_column(operator.operator2))
```
jpbrodrick89: This breaks when the in_structure of operator1 is a PyTree. Three options to fix:

1. use unravel/ravel around `operator1.mv`
2. have `first_column` return a PyTree (which is a fairly big change that probably suggests supporting PyTree's in CirculantLinearOperator and then probably in the solver too).
3. just use the generic fallback, in most cases it won't be much/any slower (but based on your PR comment it sounds like this might introduce a recursion).

1 is probably the right solution but happy to support others if you have a good reason.

lineax/_solver/circulant.py
```diff
@@ -0,0 +1,142 @@
...
+            fft_fn = jnp.fft.fft
+            ifft_fn = jnp.fft.ifft
+        else:
+            fft_fn = jnp.fft.rfft
+            ifft_fn = ft.partial(jnp.fft.irfft, n=len(vector))
+        vector_fft = fft_fn(vector)
+
+        if not self.well_posed:
+            (size,) = eigenvalues.shape
+            rcond = resolve_rcond(self.rcond, size, size, eigenvalues.dtype)
```
jpbrodrick89: Note that `size` of eigenvalues does not correspond to `len(vector)` for real. Just use `len(vector)` directly. `eigenvalues.dtype` is still correct I think.

This is a bit confusing but I think my suggestion is the right one. Happy to be proven wrong though.

jpbrodrick89:
Very nice @BalzaniEdoardo! Thanks for your patience here, we are very nearly there! Happy with all the decisions you flagged. Note that PR #221 made the same decision as you with defaulting to solve instead of lstsq so that's a sensible decision. However, we currently can't test `well_posed=False` through the `pseudoinverse=True` chain because of this same issue and I need to try and get that PR over the line so we don't have further issues. Don't let that delay the merge of this PR, but this does motivate me to try get that PR over the line.

Other than the minor new unresolved comments above I think we're good to go! Thanks again!

lineax/_operator/binary.py
BalzaniEdoardo: good catch!

BalzaniEdoardo:
@jpbrodrick89 back at it! sorry for the delay, I can allocate slot of times every few days. I should be done with the round of reviews.

Thank you for being on top of this one!

lineax/_solver/circulant.py
BalzaniEdoardo: yes, I did realize it in compute and I used `len(vector)` there, but I missed this mistake. It revealed a gap in the test coverage, so I added a test that covers this bug.

BalzaniEdoardo:
> Approved! Thanks so much for taking on board the comments and getting this to a really nice place. Really happy to have this in.
> 
> I made a small change to the assertion for the rcond in case something in the future makes this flaky. Hope that's ok.
> 
> If tests pass and you're happy with everything, lmk and we can get this merged.
> 
> Thanks again! 😃

absolutely, I learned a lot about lineax with this. very cool package

jpbrodrick89:
@ceyron hope you're well, I was wondering whether any of this infra (especially `first_column` for circulant column extraction) might be useful in exponax especially to help users easily use exponax on new equations?

Ceyron:
Thanks for tagging!

I think it might not be directly applicable, since for Exponax one would need more detailed access to the representation in Fourier space (besides just the matrix-vector product), e.g., [to evaluate the coefficients in the numerical scheme](https://github.com/Ceyron/exponax/blob/849f2c4873982ecbb95287b77dca97fe0ee80697/exponax/etdrk/_etdrk_2.py#L77-L89) or to take the elementwise exponential (to get the matrix exponential).
But as you say, `first_column` could be a nice tool to extract the equivalent circulant matrix (which likely will not have compact stencils but a dense column due to spectral differentiation) from linear time steppers or linearized ones. I will try to add this to the docs if I have some time in the future (https://github.com/Ceyron/exponax/issues/108 ).

jpbrodrick89:
@Ceyron thanks for taking a look, if by Fourier representation you just mean the eigenvalues can be obtained by Circulant().init(real_space_operator)[0] which calls first_column under the hood and then does the fft to get the eigenvalues. Is this what you were looking for?

Ceyron:
Yes, I meant the eigenvalues. Based on this, one could potentially use it to build steppers for linear PDEs directly (using the Circulant matrix to represent the spatial discretization). Though addressing nonlinear PDEs wouldn't be any easier with this than it currently is.
But since Exponax uses spectral differentiation in space, I'm unsure if there is need to use any other discrete differential operator (e.g., finite differences). Maybe for educational/research porpuses. I will give it some more thought and try to add it to the docs. 😊

